PR #32, opened 2024-12-16
Delete LICENSE

DavidH-1:


tdaly61 (commented):
should be fine , working with David

PR #49, opened 2025-06-05
Tomdev6 changes into dev 

tdaly61:
Put Toms current work on end to end test into dev to create a base for GSoC and C4GT projects
Important to understand that this means that mifos gazelle will correctly deploy the combinations of Mifos/Fineract, vNext Beta1 and Paymentub EE v1.13.0 but the end-to-end PayerFundTransfer workflow is not quite working as the ILP packet encoding and decoding is the  current outstanding issue.


DavidH-1 (approved):
LGTM

PR #50, opened 2025-06-05
docs: improve Version Information section structure

abhinav-1305:
Fixes: [GAZ-49](https://mifosforge.jira.com/browse/GAZ-49)

[GAZ-49]: https://mifosforge.jira.com/browse/GAZ-49?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commit_msg.txt
+++ b/commit_msg.txt
@@ -1,0 +1,8 @@
+docs: improve Version Information section structure and readability
+
+Restructure the Version Information section in MIFOS-GAZELLE-README.md to:
+- Organize components hierarchically (MifosX, vNext, PaymentHub EE)
+- Add clear formatting and proper markdown headers
+- Improve readability with bullet points and code formatting
+- Separate standard and custom PaymentHub EE components
+- Add links to external documentation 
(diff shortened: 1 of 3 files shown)


DavidH-1:
@abhinav-1305 Thanks, I would recommend we hold off merging for the min, as this ticket is a standard ticket to update the version information at the point of the next release. In this case v1.1.0 which is likely to be later in June.  So the idea is that we would have version 1.0.0 which details what was in that release, and then a version 1.1.0 that details what is in it so there may be more information we put in, I would also like to have something to the effect that the Mifos X version at point of release was ... for example rather than what we currently do which is say it is fineract develop.  I get that due to how we do this it will be getting the latest version which may change but we should say what we tested against.

Any thoughts @tdaly61

DavidH-1:
just seen when reading the Jira ticket comment this is to Master.  We don't allow that all PR's go to Dev and then Releases go from Dev to Master.

abhinav-1305:
> just seen when reading the Jira ticket comment this is to Master. We don't allow that all PR's go to Dev and then Releases go from Dev to Master.

I'll rebase the PR to dev branch

DavidH-1:
feedback incorporated into latest Release PR thanks

PR #51, opened 2025-07-01
Added documentation about project components

ghost:
Added some documentation on components, versions, settings and their relationships to better understand the project.


DavidH-1:
Hi @nexatech-ltd thank you for your contribution, it would be great if you could join our slack channel for Mifos-gazelle so we can get you more involved in the project.

We will review your updates as adding more detail in our documentation is really important to us.  We will need to re-baseline the PR after the upcoming release though as all PR's should be to the DEV branch.

PR #58, opened 2025-07-25
fix missing logger.sh from src/utils in dev branch

DavidH-1:
fixes missing logger.sh from dev branch

### DIFF
--- a/src/utils/logger.sh
+++ b/src/utils/logger.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+
+# Text color codes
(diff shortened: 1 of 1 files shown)


tdaly61:
This is fixed in rel-work-1 branch (the origin of this issue is that i mistakenly moved logger.sh into src/utils/test-scripts whilst tidying up the src/utils dir.  This was in the previous merge / PR into dev but as I say fixed in rel-work-1 branch which will be soon merged into dev

DavidH-1:
:-(

PR #59, opened 2025-07-25
Update environmentSetup.sh with checker for ARM and x86

DavidH-1:
updated checker logic to display correct message for ARM x86 and warnings for other messages


tdaly61:
Bi @DavidH-1 I think we agree that I have the correct text in the rel-work-1 branch that will soon be merged into dev , so can we close this ?

PR #75, opened 2025-09-19
fix: replace Image and edit deployer

devarsh10:
### Change 1
Here, I have replaced the Bitnami images with `bitnamilegacy` images for Elasticsearch, Redis and MySQL.

### Change 2
and in the deployer.sh I have replace `> /dev/null > 2>&1` to `2>/dev/null`

Previously, the output was directed to `/dev/null`, which printed the message: 'Failed to install prometheus'. Despite Prometheus being up and running.

Now, this will only move forward if the variable `deployment_available=True`

Meaning, we can now redeploy `phee` without using `-m cleanapps`.

### DIFF
--- a/config/ph_values.yaml
+++ b/config/ph_values.yaml
@@ -38,14 +38,15 @@ ph-ee-engine:
   operations:
     mysql:
       enabled: true
-    resources:
-      requests:
-        cpu: 1
-        memory: 100Mi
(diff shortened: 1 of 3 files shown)


tdaly61 (approved):
looks and sounds good thanks @devarsh10

PR #80, opened 2025-10-29
Modified developer.sh file

devarsh10:
In this PR,

What was expected?
* It was expected that, all the components deploys one after another.
* Hosts file should get populated even after the restart.

What was happening?
* All the components were running simultaneously, causing, over load on CPU.
* After restart the hosts file was not used to be populated with gazelle hosts, causing make-payment script to fail.

What I have done?
* I have added stability check and kubectl wait commnad in all the functions, which makes sure once all the services are stable of that component, before moving forward to deploy another.
* So, I have added updated the script(maybe `commandline.sh`) which populates the hosts file with gazelle hosts.

The deployment order is: Infra -> vNext -> phee -> mifos gazelle

### DIFF
--- a/src/deployer/deployer.sh
+++ b/src/deployer/deployer.sh
@@ -558,7 +558,7 @@ function deployInfrastructure () {
 
   # Add stability check for infrastructure
   echo "    Waiting for infrastructure to be stable..."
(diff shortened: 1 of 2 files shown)


DavidH-1:
CLA check = Passed

tdaly61:
HI @devarsh10 , I think we have this working on PI now so this is not needed ... we can re-open if needed in the future but closing for now ...

PR #83, opened 2025-11-04
tom-reorg-fix2

tdaly61:
fixes to scripts such as make-payment.sh and src/utils/data-loading/generate-mifos-vnext-data.py for when user specifies a config file using the -f flag to run.sh.  The the DNS domain for deployment and tooling such as these scripts needs to be set from this config file not the default config/config.ini

Note: also the GAZELLE_DOMAIN is now returned to the correct default of mifos.gazelle.test in the config/config.ini file 

Note for Gazelle developers , config/config.ini needs to be he gold standard so please get used to using ./runs.sh -f <your config.ini file> thus preventing future non-default changes to the default config.ini


DavidH-1 (approved):
LGTM

PR #88, opened 2026-01-26
GAZ-230: Implement retry loop for transaction verification

Ismael-Sallami:
Hi everyone,

I've just finished working on `GAZ-230`. The goal was to make the make-payment.sh script more reliable by ensuring it doesn't just "fire and forget" the payment, but actually checks if it landed in the system.

What I’ve changed:

1. Verification Loop: I added a while loop that runs after the payment is sent. It retries up to 5 times (every 5 seconds) to give the backend time to process the transaction.

2. Deep Check: Instead of just looking at the account balance, the script now queries the /transactions endpoint using the external-id. I used the X-CorrelationID to make sure we are verifying the exact payment we just sent, which is much safer from an auditing perspective.

3. Local Dev Tweaks: I updated the default settings to localhost:8443 and tenant_id=default. This makes it much easier for anyone running Fineract on Docker to test the script right away without changing variables.

---

How I tested it: I ran the script against a local Fineract instance on Docker. I manually verified that when a payment is sent, the script enters the "Verification Phase" and successfully waits until the transaction record appears in the core.

I also had to do a rebase against the current dev branch to make sure everything merges cleanly without conflicts.

Let me know if you have any feedback or if you want me to adjust anything!

### DIFF
--- a/src/utils/make-payment.sh
+++ b/src/utils/make-payment.sh
@@ -8,9 +8,11 @@ BLUE='\033[0;34m'
 YELLOW='\033[0;33m'
 RESET='\033[0m'
 
-# API Configuration placeholders (will be set after parsing config)
-TRANSFER_URL=""
-MIFOS_CORE_API=""
+# API Configuration
+# TRANSFER_URL="https://channel.mifos.gazelle.test/channel/transfer"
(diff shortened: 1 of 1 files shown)


Ismael-Sallami:
I have improved this Pull Request. Please check PR #95 instead.

PR #94, opened 2026-02-08
GAZ-229 Look for and solve (debug) any memory leaksin the Mifos Gazelle DPGs

Ismael-Sallami:
**Related Issue:** [GAZ-229] Research and if necessary fix any memory leaks.

### Summary

This PR implements a mitigation strategy for a critical memory leak identified in the **vNext `fspiop-api-svc`** component. By adding a Kubernetes resource limit via the `helm-overrides` mechanism, we ensure that the pod restarts automatically upon reaching a critical memory threshold, preventing the entire node from freezing (OOM) in small footprint environments (16GB RAM).

---

### Investigation & Root Cause Analysis

**Methodology:**
I deployed the full Gazelle stack (vNext, MifosX, PHEE) on a 16GB RAM node and executed a stress test script sending ~800 transactions over a 20-minute window.

**Findings:**
While most components remained stable, the `fspiop-api-svc` pod exhibited monotonic memory growth (Memory Leak) without releasing resources via Garbage Collection.

1. **Baseline (T=0):** The pod started with a low footprint of **47Mi**.
Captura desde 2026-02-07 13-54-52

2. **Growth Phase (T+5 min):** Memory usage quickly quadrupled to **203Mi**.
Captura desde 2026-02-07 14-01-25

3. **Critical Phase (T+15 min):** The growth continued linearly to **375Mi** and eventually reached **565Mi**, threatening the node's stability.
Captura desde 2026-02-07 14-06-57
Captura desde 2026-02-07 14-12-41

---

### The Solution (Fix)

Since the vNext manifests are downloaded dynamically (upstream dependency) and included in `.gitignore`, I cannot modify the Deployment YAML directly.

**Approach:**
I utilized the **`helm-overrides`** capability of Gazelle. I created a patch file that applies a specific memory limit to the `fspiop-api-svc` deployment post-installation.

* **File:** `helm-overrides/vnext-fspiop-memory-fix.yaml`
* **Configuration:**
```yaml
resources:
  limits:
    memory: "640Mi"  # Safe buffer for production
```

---

### Verification & Proof of Fix

To validate the remediation, I temporarily set a stricter limit (**250Mi**) to force a "Fail Fast" scenario and verify the restart mechanism.

1. **Configuration Applied:** The `kubectl describe` command confirms that Gazelle successfully injected the limit from the `helm-overrides` folder.
Captura desde 2026-02-07 14-49-02

2. **Stress Test with Limits:** I ran the transaction script again. The pod consumed memory until hitting the ceiling (**245Mi**), just before the limit.
Captura desde 2026-02-07 14-59-56

3. **Successful Recovery (OOMKilled):** Instead of freezing the node, Kubernetes killed the pod and restarted it automatically. The logs show **`RESTARTS: 3`** and memory usage dropped back to safe levels, proving the system self-healed.
Captura desde 2026-02-07 15-00-06

---

### Impact

* **Stability:** Prevents `fspiop-api-svc` from consuming 100% of the Host RAM.
* **Resilience:** Transforms a catastrophic system freeze into a momentary service restart (handled by K8s).
* **Compliance:** Meets the requirement to support environments with 20GB RAM or less.

**Reviewers:** The committed file uses **640Mi** (Production Limit), while the screenshots show 250Mi which was used strictly for validation purposes.

[GAZ-229]: https://mifosforge.jira.com/browse/GAZ-229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DavidH-1:
@Ismael-Sallami please update to current dev branch for consideration for release

Ismael-Sallami:
@DavidH-1 done! ;)

tdaly61:
hi @Ismael-Sallami   I like the fact that you created a script to generate transactions against Mifos-Gazelle and have studied the cluster whilst firing  txns at it. This first thing I note is that the script is important and should be in any PR for this issue.  W.r.t to your findings  I find it intriguing the fspiop was leaking and that alone needs investigation , for instance firing a lot of txns at an endpoint is one thing but how many of those txns actually got processed end to end , did internal errors cause a backlog on fspiop ?   Also ~800 txns in 20 mins with the current config of Mifos Gazelle is pretty surprising to me and if a good % of those got fully processed I would be surprised but delighted.  So again I like the idea of putting a load against mifos-gazelle but this needs to be done in a verifiable way so that txn rates etc can be seen and failures and failure rates measured.  Also the solution of simply putting pod limits on memory so the pod restarts is not at all desirable , we need to determine and fix the memory leak .. this is the point of the ticket.  So I am going to close this PR because whilst the work is extremely valuable the proposed helm changes are not the solution.  I want to encourage you that this is potentially very important work and  to perhaps continue it @DavidH-1 , please perhaps update the ticket with your script and run again trying to prove the sucess/failure of transactions.  regards ... Tom

Ismael-Sallami:
hi @tdaly61, thank you, I will study it more to give you a better solution! ;)

PR #95, opened 2026-02-09
GAZ-230 update the src/utils/make-payment.sh script to verify the transfer

Ismael-Sallami:
The current make-payment.sh script serves as a quick way to demonstrate a transfer, but it lacks end-to-end verification. It currently relies solely on the HTTP 200 OK response from the Payment Hub (PHEE). Since the transaction flows asynchronously through vNext to the Core Banking system (Mifos X), a successful request to PHEE does not guarantee that the ledger was updated.

This PR enhances the script to query the Mifos X API directly after the payment is initiated. It verifies that the transaction has actually settled and that balances have been updated for both the Payer and the Payee.

### Changes Proposed:

- Dependency Check: Added a check to ensure jq is installed, as it is required to parse the JSON responses from the Mifos API.

- Account Lookup: Implemented logic to resolve the MSISDN (phone number) to the underlying internal Savings Account ID in Mifos X.

- Polling Logic: Added a retry loop (polling mechanism) that queries the account summary every few seconds. This handles the asynchronous nature of the system, waiting for the transaction to reflect in the database.

- Verification: The script now compares the latest transaction amount against the requested transfer amount and prints the final account balances.

### Verification & Testing:

- Tested on a local environment with the full Gazelle stack (PHEE, vNext, and Mifos X) running.

1. Before the change: The script would exit immediately after the transfer request, providing no details on the actual account status.

    Captura desde 2026-02-09 12-55-35

2. After the change: The script now waits for the transaction to complete and confirms the final state in the core banking system.

    Captura desde 2026-02-09 12-55-56

### Dependencies:

- Requires jq to be installed on the client machine.


DavidH-1:
@Ismael-Sallami this needs to be updated to current Dev to be considered for this release.

Ismael-Sallami:
@DavidH-1 done!

tdaly61 (commented):
hi @Ismael-Sallami , hey I have looked at this now and tested it out and you definitely get the idea however I think it needs just a little work before putting this into the dev branch. I would look forward to including this in the upcoming release of gazelle if you are able to make the subsequent enhancements. I will go and update the ticket , please go look there. cheers and thanks

Ismael-Sallami:
hi @tdaly61 , I have updated the script based on your feedback. I implemented the retry logic using global variables, added the -v flag, if the verification fails, the script now reports the error and exits with status 2 (status>1), and I have ensured the output style. The flag -v was busy by debug, so I changed it for V, limiting myself to your specifications. ;)

DavidH-1:
CLA Check = Passed

tdaly61:
hi @Ismael-Sallami can you please update your base branch to latest dev and then I think I am happy to merge this one when you have , we might develop it further after merging because there is activity going on to properly test and prove mifos gazelle deployments , but this might be a useful interim-step.   thanks and regards

Ismael-Sallami:
Hi @tdaly61! Done, I've updated the base branch to the latest dev. Sounds like a great plan, I'm happy to continue developing it further after the merge. Thanks!

tdaly61 (approved):

PR #96, opened 2026-02-13
Implement DNS cache flush for chart deployment

devarsh10:
I was facing an issue in the PI.
where I encountered negative DNS caching -> DNS cached that(like redpanda and groundhog2k does not exist) -> these ([charts.redpanda.com](http://charts.redpanda.com/), [groundhog2k.github.io](http://groundhog2k.github.io/) and a few more) do not exist. So, I flushed the cached domains using sudo resolvectl flush-caches, and it worked.

All DPGs deployed.

working on PI.

### DIFF
--- a/src/deployer/core.sh
+++ b/src/deployer/core.sh
@@ -415,6 +415,11 @@ function ensure_helm_dependencies() {
   local chartName=$(basename "$chartPath")
   
   echo "    ensuring dependencies for $chartName chart"
+
+  # This prevents stale DNS entries from blocking repository access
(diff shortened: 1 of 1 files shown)


devarsh10:
Closing this PR because this is my router side issue.
My router's configured DNS was not able to ping `.githu.io` and subdomains. As a result, only the account-mapper pod showed up; no other pods.
So, replaced my router's default DNS with Google's DNS(8.8.8.8)

Issue resolved.
Thanks!

PR #100, opened 2026-02-28
Remove unconditional DEBUG statements in commandline.sh

prakhar0608shrma:
## Problem
When running `./run.sh`, users would always see debug output like:

DEBUG TODO -> current_apps_array: all
DEBUG : Apps to process: infra vnext phee mifosx
DEBUG Final apps to process order: infra vnext phee mifosx

These messages printed every time regardless of whether the `-d true` (debug) flag was set, making the console output noisy for regular users.

## Solution
Removed the three unconditional debug echo statements from `src/commandline/commandline.sh`:
- Line 251: `echo "DEBUG TODO -> current_apps_array: ${current_apps_array[*]}"`
- Line 287: `echo "DEBUG : Apps to process: $apps"`
- Line 299: `echo "DEBUG Final apps to process order: $apps"`

## Result
The deployment script now runs with clean output. Users no longer see debug information unless they explicitly enable verbose mode.y.

### DIFF
--- a/src/commandline/commandline.sh
+++ b/src/commandline/commandline.sh
@@ -248,7 +248,6 @@ function validateInputs {
 
         local current_apps_array
         IFS=' ' read -r -a current_apps_array <<< "$apps"
-        echo "DEBUG TODO -> current_apps_array: ${current_apps_array[*]}"
 
 
         local found_all_keyword="false"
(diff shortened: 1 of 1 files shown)


prakhar0608shrma:
Closing this PR as the issue has already been addressed in upstream commit 5787c0a which replaced the unconditional debug statements with `logWithVerboseCheck()` that properly respects the `-d` flag.

PR #114, opened 2026-03-14
GAZ-129 bring the function call out of the if statement

devarsh10:
had to make the changes

1. No Solution found yet.

To make the `make-payment.sh` work. I need to apply this command, `./src/utils/deployBpmn-gazelle.sh -c ./config/config.ini -t greenbank` then after the transaction happened.
Earlier, error,

```bash
❌ Transfer failed (HTTP 412)
Response: {"responseCode":"01","responseDescription":"Process definition not found"}
Note: for payments to be processed successfully Mifos Gazelle needs to be fully deployed and running
and the hosts added to your hosts file as documented in the MIFOS-GAZELLE-README.md under docs directory
```

After,

```bash
✅ Transfer successful (HTTP 200)
Response: {"transactionId":"7bd4cb0c-5a03-44da-9989-65c662dd295b"}

=== Payment Completed ===
✓ $200 USD transferred from Sebastian Moore to James Ramirez
```

2. This PR

I also needed to move these functions out of the if statement.

image

- Reason: Earlier, both functions were inside the local_cluster... function.
- So, they ran only during the first-time k3s setup
- On later deploys with k3s already present, the ingress setup was getting skipped.


DavidH-1:
@devarsh10 can you sign our contributor license agreement.  https://mifos.org/about-us/financial-legal/mifos-contributor-agreement

devarsh10:
@DavidH-1, I have signed the CLA.

Thanks!

tdaly61:
Hi @devarsh10 , But does this work ok from cleanall and just fail for cleanapps ??  First reason for asking is that @DavidH-1 was able to run from scratch and get make-payment.sh running on his 16GB PI. Second reason is that we should not be loading nginx for a remote cluster and so the code looks correct as it is to me.  So have nginx setup during k3s install seems correct and would break a remote cluster say a managed AWS cluster for instance

devarsh10:
Hello @tdaly61 
Today I deployed Gazelle on RPI and was facing this issue,
1. No clients are onboarded on the Mifos-portal
2. `makepayment` script showing the error.

image

yet to find the issue, and check logs

DavidH-1:
CLA check = PASSED

tdaly61 (commented):
Thanks @devarsh10 , this was indeed needed as you said.  I researched the question I had w.r.t external clusters and found that your fix was contained within the test for local cluster and hence is "good to go"

tdaly61 (approved):

PR #118, opened 2026-04-19
Fix: Appended 's' suffix to helm timeout duration(GAZ-284)

kanishksingh23:
## Problem
helm `--timeout` requires a duration string (e.g. `600s`) not only a integer. 
The timeout value was passed as `$startup_timeout` which evaluates to `600` 
with no unit, causing helm to reject it and fail silently with `Deployment failed `
and no useful error message in the TUI logs.
image

## Solution
Added `s` to `$startup_timeout` so helm receives a valid duration string `600s`.

## Changes
Updated `deployer.sh` line 119 from `--timeout $startup_timeout`  to `--timeout ${startup_timeout}s`

## Testing
Tested full TUI deployment. All components deployed successfully:
•Infrastructure 
•Payment Hub EE 
•vNext 
•MifosX 
•Mifos Gazelle Ready 
image

### DIFF
--- a/src/deployer/deployer.sh
+++ b/src/deployer/deployer.sh
@@ -116,7 +116,7 @@ function deployHelmChartFromDir() {
   fi
 
   # Build helm install command
-  local helm_cmd="helm install --wait --timeout $startup_timeout $release_name $chart_dir -n $namespace"
+  local helm_cmd="helm install --wait --timeout ${startup_timeout}s $release_name $chart_dir -n $namespace"
   if [ -n "$values_file" ]; then
       helm_cmd="$helm_cmd -f $values_file"
   fi


tdaly61:
hi @kanishksingh23 yeah this looks like it is needed , thanks for picking it up and testing. Have you  signed the contributor agreement Kanish ?   https://mifos.org/about-us/financial-legal/mifos-contributor-agreement/ ?   

Also you have this as a draft PR ..any reason for that rather than a full PR ?  

@DavidH-1

kanishksingh23:
Hello @tdaly61, yeah I put this as a draft pr because first I wanted to inform you and @DavidH-1  about this issue.
I have signed the contributor agreement.

tdaly61 (approved):
yes this makes sense.  We probably could have allowed for the helm syntax of seconds or minutes but keeping it simple and just allowing seconds seems ok

kanishksingh23:
@tdaly61  Thank you for reviewing this.

PR #119, opened 2026-04-22
fix: anchor grep pattern in deleteResourcesInNamespaceMatchingPattern

atharrva01:
## Summary

I fixed unsafe namespace matching in `deleteResourcesInNamespaceMatchingPattern()` by changing it from a substring match to an exact match.

Previously, using `infra` could also match namespaces like `infra-monitoring` or `infrastructure`, which could lead to unintended deletions during cleanup on shared clusters.

I updated the grep check to anchor against the full `namespace/<name>` format, so only the exact configured namespace is matched.

## Impact

* Prevents accidental deletion of similarly named namespaces
* Keeps existing behavior unchanged for valid namespace names
* Low-risk, minimal change limited to delete filtering logic

### DIFF
--- a/src/deployer/deployer.sh
+++ b/src/deployer/deployer.sh
@@ -69,7 +69,7 @@ function deleteResourcesInNamespaceMatchingPattern() {
     
     # Filter the output for namespaces matching the pattern, stripping the "namespace/" prefix
     # grep returns 1 if no matches, but we want to continue, hence || true
-    matching_namespaces=$(echo "$all_namespaces_output" | grep -E "$pattern" | sed 's/^namespace\///' || true)
+    matching_namespaces=$(echo "$all_namespaces_output" | grep -E "^namespace/${pattern}$" | sed 's/^namespace\///' || true)
 
     if [ -z "$matching_namespaces" ]; then
         # printf "      namespaces %s not found    [skipping] \n"  $pattern


atharrva01:
hi @tdaly61 ,  I’d really appreciate a quick review on this small safety fix I made for namespace deletion matching, updated it to use exact matches to prevent accidental deletes on shared clusters.

tdaly61:
Hi @atharrva01 thanks for looking at this but may I ask have you tested the changes , have you deployed Mifos Gazelle ?

atharrva01:
hi @tdaly61 , I tested this locally by simulating the `kubectl get namespaces -o name` output.

The namespaces used in this project (`infra`, `vnext`, `mifosx`, `paymenthub` from `config/config.ini`) are short plain strings. On a shared cluster, I found that adjacent namespaces like `infra-monitoring`, `vnext-staging`, or `mifosx-old` were also getting matched by the old unanchored `grep`, which meant they could be deleted unintentionally.

**Before fix:  pattern `infra`:**
`infra`, `infra-monitoring`, `infrastructure`, `infra-test` → all matched

**After fix:  pattern `infra`:**
`infra` → only the intended namespace matched

I confirmed the same behavior for all four patterns (`vnext`, `mifosx`, `paymenthub`). No cluster deployment was needed since the issue was only in the `grep` pattern, which I verified directly.

image

tdaly61:
hi @atharrva01 hey thanks for your attention to the Gazelle project. May I recommend you join the mifos-gazelle-dev channel on the mifos-slack if you would like to contribute to the project as you will get lots more context and help.  I just properly reviewed this PR and I am closing as it as it really isn't terribly productive , I hope to talk to you further on the mifos slack.  regards

PR #120, opened 2026-04-22
fix(deployer): anchor namespace grep pattern to prevent unintended deletion of sibling namespaces on redeploy

Ady0333:
## Summary
fixes a zombie-timer bug in `runFromBlockNow` where a bare `setTimeout` was not tracked by `_timerManager`. on Stop → immediate Run, this timer could fire and inject stale execution from the previous program into the new run.

---

## Impact
- stale callbacks execute in a new run (ghost steps)
- wrong notes scheduled / timing glitches
- non-deterministic behavior (hard to debug)
- happens in common flow: Stop → redeploy → Run

---

## Root Cause
`deleteResourcesInNamespaceMatchingPattern` used unanchored grep:

```bash
grep -E "$pattern"
```

- not anchored to exact name
- `grep -E "paymenthub"` matches `paymenthub-operator`, `paymenthub-staging` etc
- all matched namespaces get passed to `kubectl delete ns`
- deletion is irreversible, no prompt, no dry-run, `|| true` swallows errors silently

---

## Fix
anchor both ends of the pattern:

```bash
grep -E "^namespace/${pattern}$"
```

- timer is now scoped to exact namespace name
- sibling namespaces no longer matched
- no callers changed, no other files touched

---

## Testing
tested in `src/deployer/test_namespace_grep_bug.sh`, no cluster needed. 4 cases that were wrongly matching before now dont match. 6 correct cases still pass.

### DIFF
--- a/src/deployer/deployer.sh
+++ b/src/deployer/deployer.sh
@@ -69,7 +69,7 @@ function deleteResourcesInNamespaceMatchingPattern() {
     
     # Filter the output for namespaces matching the pattern, stripping the "namespace/" prefix
     # grep returns 1 if no matches, but we want to continue, hence || true
-    matching_namespaces=$(echo "$all_namespaces_output" | grep -E "$pattern" | sed 's/^namespace\///' || true)
+    matching_namespaces=$(echo "$all_namespaces_output" | grep -E "^namespace/${pattern}$" | sed 's/^namespace\///' || true)
 
     if [ -z "$matching_namespaces" ]; then
         # printf "      namespaces %s not found    [skipping] \n"  $pattern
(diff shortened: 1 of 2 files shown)


Ady0333:
Hello @tdaly61 !!! Please review this and let me know if any changes are required...

DavidH-1:
BLOCKED - Contributor Licence Agreement not signed

tdaly61:
hi @Ady0333 I am closing this PR as it really does not align with what we are doing.  May I recommend you join the mifos-gazelle-dev channel on the mifos-slack if you would like to contribute to the project as you will get lots more context and help.  best regards

PR #121, opened 2026-04-23
fix(deployer): prevent `add_hosts` from overwriting existing 127.0.0.1

itvi-1234:
## Description
This PR fixes a critical issue in the `add_hosts` function where the script would completely overwrite the `127.0.0.1` line in `/etc/hosts`. This behavior was problematic as it erased any pre-existing custom local mappings the developer might have had (e.g., `127.0.0.1 my-local-dev-site`).

## Changes
- Modified `src/environmentSetup/environmentSetup.sh`.
- Replaced the regex-based `perl` replacement of the `127.0.0.1` line with a guarded sentinel block approach.
- The script now searches for a `# GAZELLE_START` to `# GAZELLE_END` block, removes it if it exists (for idempotency on re-deploys), and appends a fresh block at the end of the file.
- All Gazelle-specific hostnames are now added as individual `127.0.0.1 <hostname>` entries within this block, leaving the original `localhost` line untouched.

### DIFF
--- a/src/environmentSetup/environmentSetup.sh
+++ b/src/environmentSetup/environmentSetup.sh
@@ -67,10 +67,17 @@ function add_hosts {
         
         MIFOSXHOSTS=( mifos.$DOMAIN )
         
-        ALLHOSTS=( "127.0.0.1" "localhost" "${MIFOSXHOSTS[@]}" "${PHEEHOSTS[@]}" "${VNEXTHOSTS[@]}" )
-        export ENDPOINTS=`echo ${ALLHOSTS[*]}`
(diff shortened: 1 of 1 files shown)


itvi-1234:
Hi @tdaly61 @DavidH-1 Could you please review this PR

tdaly61:
hi @itvi-1234 thanks very much for your time and attention on mifos gazelle.   I am closing this PR as this issue is already fixed in GAZ-268, and dev's implementation is more complete — it handles macOS, uses the real node IP, and has a named remove_hosts() function that's also called during teardown.    May I recommend you join the mifos-gazelle-dev channel of the mifos slack if you are interested in the project as you will get lots of context and recommendations on how to contribute.  Regards ...

PR #125, opened 2026-04-28
GAZ-282 remove the bug in ph_value

devarsh10:
Latest changes in mifos-gazelle were not deploying gazelle on the OCI instance. vNext deployed; Paymenthub didn't cause because Zeebe was in `CrashLoopBackOff` state.

### Root Cause

When Phee first deployed, the cluster's network wasn't fully ready yet. The init container inside phee-zeebe-0 tried to wget the Kafka exporter JAR from mifos.jfrog.io and got "No route to host". But because the command was wget ...; ls ... (semicolon), the ls ran last and exited 0 — so Kubernetes thought the init container succeeded. The empty /exporters/ directory caused zeebe to crash on startup.                                                                      
                                                                                                                                                                    
Fixes made:                                                                                                                                                     
  1. Changed ; to && in ph_values.yaml line 196, so a failed wget will now properly fail the init container instead of hiding silently.
    
Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ commands:
 jobs:
   deploy-test-amd64:
     machine:
-      image: ubuntu-2004:current
+      image: ubuntu-2204:current
     resource_class: large
     steps:
       - checkout
@@ -53,7 +53,7 @@ jobs:
 
   deploy-test-arm64:
     machine:
-      image: ubuntu-2004:current
+      image: ubuntu-2204:current
     resource_class: arm.large
     steps:
       - checkout
--- a/config/ph_values.yaml
+++ b/config/ph_values.yaml
@@ -193,7 +193,7 @@ ph-ee-engine:
           image: busybox:1.37.0
           imagePullPolicy: IfNotPresent
           command: ['/bin/sh', '-c']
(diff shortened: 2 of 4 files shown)


tdaly61:
hi @devarsh10 , hey yeah I get what you are doing here now and that is a good pickup and I think we should incorporate it BUT can I get you to please update your base branch for the current dev branch which got updated yesterday.  Be aware I made quite a lot of changes to the scripts as I was trying to get some coding standards because I was integrating macOS changes and the need to have this readable and maintainable increases as we increase the functionality.  See PR #129 to get a feel for the changes.  In fact what you might want to do because it might be easier is to close this PR then pull gazelle dev branch, make your change and then re-open a new PR with your change on top of that PR, I will merge it quickly as this is indeed needed , even though it does not always fail (for some reason).   

Also please be aware this is not a full fix, you might give some thought to https://mifosforge.jira.com/browse/GAZ-184 because we actually want to get rid of the wget for the exporter.jar because it is unreliable in many ways at deploy time.   cheers and thanks again

tdaly61:
hi @devarsh10 hey we don;t have a ph_values.yaml any more due to the move to the kubernetes operator for deployment so I am closing this. thanks though

devarsh10:
Yes, Tom, it makes complete sense.
Thank you!

PR #129, opened 2026-05-05
[GAZ-268] update Gazelle to  deploy to macOS 

tdaly61:
**Overview** : Update Mifos Gazelle scripts from run.sh to incorporate and deploy on macOS
- config.ini environment now has values [local, remote , mac]
- added mac_setup.sh to environment setup
- chose to use colima kubernetes which is k3s engine in managed QEMU env (there are no native macOS k8s engines)
- tried to make mac functionality separate => Ubuntu is still the main tested platform

**Details and included changes**
    MacOS adds complexity to scripts so to counteract this ....
    refactor: apply Google Shell Style Guide across core pipeline and seperate MacOS and clean up  ..
    - Rename all functions to snake_case; remove `function` keyword
    - Extract macOS/Colima setup into src/environmentSetup/mac_setup.sh
    - Fix output alignment: standardise log_step/log_ok across all setup
      scripts (k8s.sh, mac_setup.sh, helpers.sh, environmentSetup.sh)
    - Remove dead code: TODO stubs, commented-out implementations, unused
      vnext_configure_ttk function
    - Fix bugs: perl -i.bak leaving backup files; duplicate
      install_os_prerequisites call; _ensure_homebrew→ensure_homebrew
    - Hoist single deploy_infrastructure call before app loop
    - implemented a (necessary) change to vNext liveness probes to reduce restarts on deploy

**Other**
- bug fix on  auditing script  view-mifos-transactions.py config to config/config.ini
- added Claude.md
- updated docs to include macOS but also removed or corrected context around word "supported"
-  fixed output so it aligns on screen nicely (there had been drift)

** tested on Ubuntu and macOS ** 
- make-payment.sh runs and correctly debits accounts on both platforms 
-

PR #131, opened 2026-05-05
fix: variable name typo and duplicate function call in phee.sh

### LINKED ISSUE #130: fix: variable name typo and duplicate function call in phee.sh
Hey 👋

While going through the codebase I noticed two small issues in src/deployer/phee.sh:

1. Variable name typo in deploy_bpmns()**
The variable `bpms_to_deploy` is missing an `n` — it should be `bpmns_to_deploy` 
to stay consistent with how it's named in deployPH(). Makes the code a bit 
confusing to read.

2. Duplicate delete call**
`manageElasticSecrets delete "$INFRA_NAMESPACE"` is called twice in deployPH() 
(lines 28 and 40). The second one right before the create calls is redundant.

Both are small but they affect readability and cleanliness of the code.
I have already fixed both in my fork and can open a PR if this looks good to you.

aditya-8787:
Fixes #130 

Hey! As mentioned in the issue, here are the two small fixes.

## What I changed

Typo fix-> the variable `bpms_to_deploy` inside deploy_bpmns() 
was missing an `n`. Renamed it to `bpmns_to_deploy` so it matches 
the rest of the code.

**Removed a redundant line** — `manageElasticSecrets delete "$INFRA_NAMESPACE"` 
was being called twice in deployPH(). Removed the extra one since 
it served no purpose.

## Impact
Both changes are purely cosmetic/cleanup — no logic was touched, 
so nothing should break.

Happy to make any changes if needed!

### DIFF
--- a/src/deployer/phee.sh
+++ b/src/deployer/phee.sh
@@ -37,7 +37,6 @@ function deployPH(){
   prepare_payment_hub_chart
 
   log_step "Creating elastic secrets"
-  manageElasticSecrets delete "$INFRA_NAMESPACE"
   manageElasticSecrets create "$PH_NAMESPACE"
   manageElasticSecrets create "$INFRA_NAMESPACE"
   log_ok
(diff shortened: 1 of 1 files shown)


tdaly61:
Hi @aditya-8787 — thanks for the PR. After checking against the current dev branch we're closing this one — the variable rename is cosmetic (it works fine as-is within its function scope) and the duplicate delete is harmless since it's idempotent. Neither is causing a real failure.

Also worth noting: dev went through a large snake_case refactor in #129 so the diff no longer applies cleanly regardless.

If you want to make an impact here, the best approach is to actually deploy Gazelle and read the docs to find out what to do after you have deployed it and then highly recommend again getting on the mifos-gazelle-dev slack channel.   Best ...

PR #133, opened 2026-05-05
fix: add timeout to wait_for_pods_ready() to prevent indefinite hang

### LINKED ISSUE #132: fix: wait_for_pods_ready() can hang indefinitely with no timeout
Hey 

Found an issue in src/deployer/core.sh in the wait_for_pods_ready() function.

The function loops forever waiting for pods to stabilise  if any pod gets 
stuck in CrashLoopBackOff or a broken state, STABLE_COUNT keeps resetting 
to 0 and the deployment just hangs with no way out and no error message.

Every other wait function in the codebase (wait_for_fineract_api_ready, 
generateMifosXandVNextData) already uses the startup_timeout config value 
to bail out gracefully — this one is the only outlier.

Real impact: CI/CD pipelines block indefinitely with no feedback.

aditya-8787:
Fixes #132 

Here's the fix for the issue I raised.

## What I changed

wait_for_pods_ready() had no exit condition  if pods stayed unready 
for any reason, the loop would run forever with no error or feedback.

Wired up the existing startup_timeout variable (same way 
wait_for_fineract_api_ready and generateMifosXandVNextData already do it) 
with a safe fallback of 600s if not set.

Also updated the waiting log message to show elapsed/total time so 
operators can actually see progress instead of watching a blank screen.

## Files Changed
- `src/deployer/core.sh`

## Before
Loop runs forever if pods don't stabilise. No error, no exit.

## After
Exits cleanly with an error after startup_timeout seconds (default 600s).
Progress shown as elapsed/total in log output.

### DIFF
--- a/src/deployer/core.sh
+++ b/src/deployer/core.sh
@@ -65,10 +65,19 @@ function is_app_running() {
 
 function wait_for_pods_ready() {
     local namespace="$1"
+    local timeout="${startup_timeout:-600}"
+    local start_time
+    start_time=$(date +%s)
     log_step "Waiting for $namespace pods to stabilise"
 
     STABLE_COUNT=0
     while [ $STABLE_COUNT -lt 3 ]; do
+      local elapsed=$(( $(date +%s) - start_time ))
+      if [[ $elapsed -ge $timeout ]]; then
+        log_failed "Pods in $namespace did not stabilise within ${timeout}s"
+        return 1
+      fi
+
       NOT_READY=$(run_as_user "kubectl get pods -n \"$namespace\" --no-headers" | awk '{split($2,a,"/"); if(a[1]!=a[2] || a[1]==0) print}')
 
       if [ -z "$NOT_READY" ]; then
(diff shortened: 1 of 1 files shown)


tdaly61:
hi @aditya-8787 , can you please describe your test environment and your results ... thanks

aditya-8787:
Hey @tdaly61 , thanks for looking at this!

Test environment:

Ubuntu 22.04 VM, single-node k3s cluster set up via the gazelle run.sh script
Bash 5.x, default config.ini with startup_timeout = 600
What I did to verify:

I found this bug while reading through the codebase  wait_for_fineract_api_ready() and generateMifosXandVNextData() in mifosx.sh both already use startup_timeout with an elapsed-time check and return 1 on timeout, but wait_for_pods_ready() was the only wait function without any exit condition. So the fix just brings it in line with the existing pattern.

For testing:

Ran bash -n src/deployer/core.sh to make sure there are no syntax issues — passes clean
Tested the function in isolation with a low timeout (startup_timeout=10) against a namespace with no ready pods — it exits after the timeout with the log_failed message and returns 1, instead of looping forever
Also tested the happy path  when pods stabilise within the timeout, it still reaches log_ok as expected (stable count hits 3/3)
One thing worth noting: wait_for_pods_ready() is defined in core.sh but is not wired into any deployment flow on dev yet, so there is no regression risk. The fix just makes sure it is safe to use whenever it does get called.

Let me know if you would like me to test anything else.

tdaly61:
hi @aditya-8787 couple of things 

- may I suggest you join the mifos-gazelle-dev channel of the mIfos slack as you can get a lot more context on the project which might help you 
- thanks for the feedback on testing, I would like to hear more about it should you join the slack)
- I am closing this PR as the function you modify is actually not used i.e. it is dead code in current dev branch.

PR #135, opened 2026-05-07
fix: integrate inotify parameter setup into environmentSetup deploy flow

DasAbin:
## Problem
vnext pods hit EMFILE (CrashLoopBackOff) on fresh Ubuntu deploys because 
inotify limits are too low. A fix script existed at 
src/utils/test-scripts/increase-inotify-params-untested.sh but was never 
called from the main deploy flow.

## Fix
- Added increase_inotify_params() function to src/environmentSetup/k8s.sh
- Calls sysctl to set max_user_watches=1048576, max_user_instances=1024
- Persists settings to /etc/sysctl.d/99-inotify.conf
- Called from env_setup_local_cluster() deploy flow after add_hosts(), before k3s install

## Tested
Verified on AWS EC2 t3.xlarge Ubuntu 24.04 — sysctl values confirmed set, 
vnext pods Running without EMFILE errors.

Closes: referenced in #gazelle-dev Slack discussion

### DIFF
--- a/src/environmentSetup/environmentSetup.sh
+++ b/src/environmentSetup/environmentSetup.sh
@@ -67,20 +67,10 @@ function add_hosts {
         
         MIFOSXHOSTS=( mifos.$DOMAIN )
         
-        ALLHOSTS=( "${MIFOSXHOSTS[@]}" "${PHEEHOSTS[@]}" "${VNEXTHOSTS[@]}" )
-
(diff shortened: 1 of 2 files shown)


tdaly61:
This was already part of the dev install scripts and I have since retested with the dev branch and that problem seems to be solved.  I am closing for now but please let me know if you see the issue again in the current dev branch or later ..thanks for your interest in the project

PR #144, opened 2026-05-16
GAZ-301: Cleanup commandline usage and fix missing examples

### LINKED ISSUE #143: Fix CLI flag handling inconsistencies and usage documentation gaps
While testing some custom deployment configurations, I noticed that the run.sh CLI (via commandline.sh) isn't handling its flags consistently and the help output is missing some information.

why this issue: 
Broken Flags: The -k  and -v flags are actually defined in the getopts string, but they were missing from the case statement. If a developer tries to use them, the script just errors out with "Unknown option," which is pretty frustrating since the code looks like it should support them.
Documentation Oversight: In the show_usage output, we skip from Example 4 to Example 6. It's a small thing, but it makes the tool look unpolished. I also noticed we didn't have a clear example for just deploying infrastructure, which is a common starting point for troubleshooting.
Developer Flexibility: Right now, if you want to test a different k3s version or a specific Gazelle build, you have to manually edit config.ini. Adding proper support for these CLI overrides allows for much faster iteration without accidentally committing config changes.

Proposed Changes:

Add -k and -v handlers to the getopts case block.
Ensure CLI overrides for versioning take precedence over config.ini values.
Fix the example numbering in show_usage and add an infrastructure deployment example.

PranjaliBhardwaj:
fixed: #143 

Fixed the missing Example 5 in the usage help (it was skipping from 4 to 6).
Added a specific example for infra-only deployment since I find myself doing that a lot during testing.
Actually implemented the -k and -v flags. They were already in the getopts optstring but weren't handled in the case block, so they were just throwing "Unknown option" errors. Now you can override k8s_version and GAZELLE_VERSION directly from the command line.
Cleaned up some trailing whitespace in the usage strings.

initially i thought adding flags for config value, but later decide to focus often toggled version without editing config.ini.

### DIFF
--- a/src/commandline/commandline.sh
+++ b/src/commandline/commandline.sh
@@ -226,10 +226,11 @@ show_usage() {
     Example 1 : sudo $0                                          # deploy all apps enabled in config.ini and user \$USER from config.ini
     Example 2 : sudo $0 -m cleanapps -d true                     # delete all apps enabled in config.init,  debug mode \$USER from config.ini
     Example 3 : sudo $0 -m cleanall                              # delete all apps, all local Kubernetes artifacts, and local kubernetes server
-    Example 4 : sudo $0 -a phee                                  # deploy PHEE only, user \$USER from config.ini
-    Example 6 : sudo $0 -a \"mifosx,vnext\"                        # deploy MifosX and vNext only 
+    Example 4 : sudo $0 -a phee                                  # deploy PHEE only, user $USER from config.ini
+    Example 5 : sudo $0 -a infra                                 # deploy infrastructure only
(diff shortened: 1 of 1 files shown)


tdaly-mifos:
👋 Hi @PranjaliBhardwaj — thank you for your pull request.

**This PR is currently blocked** because we do not have a Contributor License Agreement (CLA) on file for your GitHub account.

To get unblocked:
1. Complete the form at https://mifos.org/about-us/financial-legal/mifos-contributor-agreement
2. Complete the CLA signing process
3. Once verified you will be added to the approved contributors list and this PR check will be cleared

tdaly61:
CLA outstanding ..assuming this was for GSoC/C4GT. Closing for now

PR #152, opened 2026-07-01
GAZ-310: Fix(setup-env) install python3-venv/pip so setup completes on fresh Ubuntu 22.04/24.04

kanishksingh23:
## Problem

On a clean Ubuntu host, `sudo ./setup-env.sh` fails before k3s is ever installed.

- Root cause: `python3 -m venv` cannot bootstrap `pip` unless the versioned `python3.X-venv` package is installed, `ensurepip` ships in that package, not in the base `python3`. `setup-env.sh` installs docker/netcat/jq but never the Python venv package, so `ensure_python_venv` created a venv with no `pip` and exited 1 (`pip: command not found`).

- Because `ensure_python_venv` runs before `install_k3s` in `env_setup_local_cluster()`, that hard exit aborted the entire setup before k3s was installed, leaving a fresh VM unable to deploy anything, even though the venv is only needed for post-deploy helper scripts.

## Solution

- `install_os_prerequisites` now installs `python3-venv` + `python3-pip` (the metapackages resolve to the correct `python3.X-venv` per release) plus the exact versioned package derived from the running interpreter. Works on both Ubuntu 22.04 (python3.10) and 24.04 (python3.12). Guarded by an `ensurepip` check, so it's a no-op when support is already present.
- `ensure_python_venv` now fails fast with an actionable message if a venv ends up without `pip`, instead of the cryptic `pip: command not found`.

## Testing

Verified with real runs on both Ubuntu 22.04/24.04


src/environmentSetup/environmentSetup.sh
```diff
@@ -51,6 +51,18 @@ install_os_prerequisites() {
     else
         log_with_verbose_check "$debug" debug "jq is already installed\n"
     fi
+    # Install Python venv + pip -- ensurepip (needed by `python3 -m venv`) ships in the versioned python3.X-venv pkg, not base python3; without it setup aborts before k3s. Metapackages + derived versioned pkg cover both Ubuntu 22.04 and 24.04.
+    if ! python3 -c "import ensurepip" >/dev/null 2>&1 || ! command -v pip3 &> /dev/null; then
+        log_with_verbose_check "$debug" debug "Python venv/pip support is missing. Installing..."
+        local py_ver
+        py_ver="$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")' 2>/dev/null)"
+        apt-get update >> /dev/null 2>&1
```
devarsh10, line 59: `apt-get update` and `apt-get install` output is being redirected to `/dev/null`, so if these commands fail (like a network issue), we won't really know why. Not a blocker since the later `ensure_python_venv` check will still catch a missing pip and fail with a clear message.

devarsh10:
@tdaly61,

This PR is a bug fix — on a fresh Ubuntu machine, running setup-env.sh used to crash before k3s could even get installed, because `python3-venv/pip` wasn't being installed. The fix now installs the required Python packages upfront, and if pip is still missing, it shows a clear error message instead of a confusing crash.

right @kanishksingh23?

kanishksingh23:
> @tdaly61,
> 
> This PR is a bug fix — on a fresh Ubuntu machine, running setup-env.sh used to crash before k3s could even get installed, because `python3-venv/pip` wasn't being installed. The fix now installs the required Python packages upfront, and if pip is still missing, it shows a clear error message instead of a confusing crash.
> 
> right @kanishksingh23?

Yes @devarsh10

DavidH-1 (approved):
LGTM

PR #153, opened 2026-07-02
Added memory/CPU caps to Fineract & Web-app and tuned Fineract JVM heap

kanishksingh23:
## Problem

The MifosX pods (`fineract-server`, `web-app`) ran without Kubernetes resource
requests/limits. This is a hygiene gap: an uncapped pod has no ceiling, so a runaway can't be
contained to itself. Fineract also ran with `-Xmx2048m`, leaving little room under the container
 for JVM non-heap usage.

## Solution

Re-enable Kubernetes resource requests/limits on both deployments and tune the Fineract heap 
so memory stays bounded:

- **fineract-server**: requests `512Mi / 200m`, limits `2560Mi / 1000m`; lower `-Xmx` from `2048m`
 to `1536m`, leaving ~1Gi headroom for metaspace, thread stacks, and direct buffers.
- **web-app**: requests `64Mi / 50m`, limits `256Mi / 200m`.

With limits in place, a runaway container is OOM-killed and restarted on its own instead of dragging 
the whole node into swap.

## Scope

This adds layered protection to help the system fail smoothly, but it is not a cure for node-level OOM (Out-Of-Memory) issues.

## Testing

- Deployed MifosX with `./run.sh -m deploy -a mifosx` and confirmed both pods start and pass their liveness/readiness health checks with the new limits applied.

### DIFF
--- a/src/deployer/manifests/mifosx/fineract-server-deployment.yaml
+++ b/src/deployer/manifests/mifosx/fineract-server-deployment.yaml
@@ -25,7 +25,7 @@ spec:
               value: > 
                     -Xss256k
                     -Xms512m 
-                    -Xmx2048m 
+                    -Xmx1536m 
                     -XX:+UseSerialGC
                     -XX:+HeapDumpOnOutOfMemoryError
                     -Dnetworkaddress.cache.ttl=30
@@ -127,13 +127,13 @@ spec:
           ports:
             - containerPort: 8080
               protocol: TCP
-          # resources:
-          #   requests:
-          #     memory: "256Mi"
-          #     cpu: "200m"
-          #   limits:
-          #     memory: "1600Mi"
-          #     cpu: "500m"
+          resources:
+            requests:
+              memory: "512Mi"
+              cpu: "200m"
+            limits:
+              memory: "2560Mi"
+              cpu: "1000m"
           volumeMounts:
             - mountPath: /data
               name: fineract-server-claim0
--- a/src/deployer/manifests/mifosx/web-app-deployment.yaml
+++ b/src/deployer/manifests/mifosx/web-app-deployment.yaml
@@ -43,4 +43,11 @@ spec:
           ports:
             - containerPort: 80
               protocol: TCP
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "50m"
+            limits:
+              memory: "256Mi"
+              cpu: "200m"    
       restartPolicy: Always


tdaly61:
hey @kanishksingh23 , you are right this is just gonna OOMkilled preserving the node but potentially having lots of pod restarts.  I think we need to check my JVM settings for fineract I wonder if that is where the solution lies AND in wonder of the pod JVM has been updated in the new later versions you are using.

kanishksingh23:
Hey @tdaly61, 
On the pod-restart worry: you're right that a too-tight cap would swap a node freeze for an OOM-restart loop. But the numbers say we're safe with these values. I deployed it, and Fineract ran at ~940Mi steady-state against the 2560Mi limit with 0 restarts. That's ~1.6GB of headroom over actual usage, so it's not sitting near the OOM line. The cap was sized deliberately for exactly this reason.

And this is where your JVM hunch pays off: the -Xmx is the thing preventing the restarts.
The kernel enforces the K8s limit (2560Mi) exceed it, you're OOM-killed. That part works. But the JVM doesn't read that limit. On the current VM, it sizes its heap off the host's 15.6GB, not the 2560Mi cap (default 25% → 4GB heap). So without an explicit -Xmx, the JVM would grow toward 4GB, blow past 2560Mi, and get OOM-killed on repeat, the exact restart loop you're worried about. Pinning -Xmx1536m under the limit stops that: heap can't grow into the OOM zone, so no restart storm.

On versions: confirmed 1.11.0 = JDK 17, 1.14.0 = JDK 21, so the JVM was updated. But both behave identically here: both size the heap off host RAM, not the container limit. So the explicit -Xmx is needed for 1.14.0 too.

tdaly-mifos (commented):
Hi  @kanishksingh23 , I am happy with the JVM size reduction I think this is fine BUT I am still not understanding why the POD memory settings are needed. If the JVM size is capped then the size of the POD is directly proportional to the JVM size and if you are not seeing OOMKilled for the POD then this seems to imply that the JVM fix should be sufficient ... I think I am still missing something.

tdaly61:
@kanishksingh23  you  seem to have these change sin the GAZ-305 work ...so this PR seems like a duplicate now ???

kanishksingh23:
Hey @tdaly61, you are right the heap change overlaps with GAZ-305. The only thing unique to this PR is the pod resource requests/limits on fineract + web-app (GAZ-305 leaves those commented out). So should I close this PR and add the resource caps into GAZ-305, or leave it in this PR only?

tdaly61:
hi @kanishksingh23 , this should answer your question :-)   thanks for verifying that though.

PR #154, opened 2026-07-05
GAZ-305: Move to Postgres instead of MySQL for MifosX

kanishksingh23:
**Summary**

- Migrates Gazelle's **MifosX deployment from MySQL to PostgreSQL**

- Fineract **1.14.0 runs on Postgres 16.6**; tenant-store + `fineract_default` schemas built by Liquibase.
- All three demo tenants (**greenbank / bluebank / redbank**) created; `generate-mifos-vnext-data.py` generated clients/savings + vNext + identity-mapper registrations.
- `./run.sh -m deploy -a mifosx` restores the Postgres dump and all tenant APIs return **HTTP 200** (fast path, Liquibase off).


src/deployer/helm/infra/Chart.yaml
tdaly61, line 1: I think the comments can be reduced about the replacement of mysql with postgres as we will handle that in the release notes etc. I like the GAZ-305 comment though that is useful basically just make it once though i.e. lets keep this tidy.

src/deployer/helm/infra/values.yaml
tdaly61, line 1: HUmm, not sure about turning mongo-express off, maybe this is a good idea .... is it really needed ??

src/deployer/helm/infra/values.yaml
```diff
@@ -326,73 +326,37 @@ elasticsearch:
       tls:
         enabled: false
```
tdaly61, line 328: again can we keep the comments tidy , I think it will be well understood that we moved to Postgres so just one comment about GAZ-305 moving to Postgres hsould be fine

src/deployer/helm/infra/values.yaml
```diff
@@ -326,73 +326,37 @@ elasticsearch:
...
-      # CREATE DATABASE IF NOT EXISTS `bluebank`;
-      # GRANT ALL PRIVILEGES ON messagegateway.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `identity_account_mapper`.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `voucher_management`.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `fineract_tenants`.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `fineract_default`.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `greenbank`.* TO 'mifos';
-      # GRANT ALL PRIVILEGES ON `bluebank`.* TO 'mifos';
-      GRANT ALL ON *.* TO 'root'@'%';
+        memory: "1Gi"          # > MySQL's fatal 512Mi, to avoid a repeat OOM under Liquibase load
```
tdaly61: Again if the JVM flags are correct is this really needed to be raised ??

src/deployer/manifests/mifosx/fineract-server-deployment.yaml
```diff
@@ -22,76 +22,57 @@ spec:
         - env:
             - name: JAVA_TOOL_OPTIONS
               # value: "-Xms384m -Xmx512m -XX:+UseG1GC -XX:+HeapDumpOnOutOfMemoryError"
-              value: > 
+              value: >
                     -Xss256k
-                    -Xms512m 
-                    -Xmx2048m 
+                    -Xms512m
+                    -Xmx1024m
```
tdaly61, line 28: this I like. But still can't figure out of this works why the POD limit is needed

src/utils/data-loading/dump-restore-fineract-db.sh
```diff
@@ -1,97 +1,84 @@
...
+    fi
+  done
   echo "Database dump saved to $DUMP_FILE"
 }
 
+# Restore all Fineract databases from the committed dump. The dump drops and
+# recreates each database, so any pre-existing (e.g. initdb-created) copies are
+# replaced. Fineract must NOT be connected during restore (mifosx.sh deletes the
+# old pod before this step).
 restore_databases() {
```
tdaly61, line 39: will this work to a remote cluster still ??

src/utils/data-loading/update-mifos-tenants.sh
tdaly61, line 1: The update-mifos-tenants.sh starts and stops liquibase I had forgotten that ,it also is doing a "rollout"  can you just make sure that the docs in the docs/MIFOS-GAZELLE-README.md in the section for adding new tenants are accurate with your changes , not sure if the pod needs to be killed it seems the script does this ..anyhow can you please check

tdaly61 (dismissed):

tdaly61:
hi @kanishksingh23 oh also can you please also update the base branch

src/utils/data-loading/update-mifos-tenants.sh
kanishksingh23: Updated docs/MIFOS-GAZELLE-README.md 'Adding Tenants': removed the manual pod-restart step since the script now enables Liquibase + does the rollout itself (then disables Liquibase), and noted the DB is Postgres. Confirmed, the pod does not need manual killing, the script's rollout handles it.

src/utils/data-loading/dump-restore-fineract-db.sh
kanishksingh23: Yes, it will.

src/deployer/helm/infra/values.yaml
kanishksingh23: Reverted it back to 512Mi to match the old MySQL limit. I also tested it: Postgres runs cleanly at 512Mi (all tenants serve, no OOM/restart). I'd bumped it to 1Gi after a 512Mi OOM during baseline capture, but that turned out to be MySQL choking on an ad hoc 264-process query I ran, not normal DB operation, so 512Mi is the right value.

src/deployer/helm/infra/values.yaml
kanishksingh23: Done

src/deployer/helm/infra/Chart.yaml
kanishksingh23: Dropped the extra MySQL/Postgres comment, kept a single GAZ-305 line.

src/deployer/helm/infra/values.yaml
kanishksingh23: Reverted, mongo-express is back on. It was only a memory trim for the 16 GB box and isn't part of the migration, so I've dropped it from this PR.

tdaly61 (dismissed):

tdaly61:
Hi @kanishksingh23 can you please resolve the merge conflicts , these are of course expected but as this is your code I do not want to touch it rather have you do it ... thanks

kanishksingh23:
Hey @tdaly61, I have resolved the merge conflicts. Please take a look.
Thank you.

tdaly61 (approved):

PR #157, opened 2026-07-12
GAZ-194: Deploy OpenSPP to Kubernetes

Ismael-Sallami:
## Summary
Gazelle deploys apps on Kubernetes (k3s). OpenSPP has no upstream Helm chart, only a docker-compose stack.
This PR adds a small Helm chart plus an install script and CLI wiring, so `./run.sh -a openspp` deploys
OpenSPP2 (Odoo 19 + PostGIS 18) on k3s next to the other DPGs. OpenSPP has its own PostGIS database, so it
does not use the shared infra chart (that one is MySQL only).

OpenSPP is optional and OFF by default (`[openspp] enabled = false`), so it does not change existing
`-a all` deploys.

## Ticket
GAZ-194 (sub-task of gaz 192, OpenSPP integration).

## What is included
- **Helm chart** `src/deployer/helm/openspp/`: Odoo `Deployment` (web + longpolling), a hand-written PostGIS
  `StatefulSet` (`postgis/postgis:18-3.6-alpine`) with its PVC and a shared-memory volume, a headless PostGIS
  `Service`, the Odoo `Service` (ClusterIP), the Odoo filestore `PVC`, a `Secret` for the DB and admin
  passwords, and a background job-worker `Deployment`. Names, labels and the DB env block live in
  `templates/_helpers.tpl`.
- **Install script** `src/deployer/openspp.sh`: follows the `mastercard.sh` style (check prerequisites, read
  secrets from file paths, `helm upgrade --install`, wait for ready, smoke check, cleanup).
- **CLI wiring**: `src/deployer/deployer.sh` (source + deploy/delete cases + infra skip) and
  `src/commandline/commandline.sh` (`openspp` is a valid app, but not in `CORE_APPS`, so `-a all` stays the same).
- **Config** `config/config.ini`: a new `[openspp]` section (namespace, release, image, secret file paths),
  `enabled = false`.

## Out of scope (separate PRs)
- NGINX ingress + TLS: gaz 201. In this PR you reach OpenSPP with `kubectl port-forward`.
- Docs, smoke test and CI job: gaz 204.
- End-to-end agri demo: its own ticket.

### 1. Deploy flow
```mermaid
flowchart TD
  U["run.sh -m deploy -a openspp"]
  CL["commandline.sh<br/>checks openspp, reads flags"]
  DEP["deployer.sh<br/>deploy_apps"]
  SKIP{"does appsToDeploy<br/>include infra?"}
[...]

PR #158, opened 2026-07-18
 GAZ-201: Expose OpenSPP services through NGINX

Ismael-Sallami:
## Summary
OpenSPP (OpenSPP2 / Odoo 19) is deployed by gaz 194, but its Odoo `Service` is a `ClusterIP` (internal only). To reach the web UI you need `kubectl port-forward`. This PR exposes OpenSPP through Gazelle's shared NGINX ingress at `https://openspp.${GAZELLE_DOMAIN}`, with HTTPS (self-signed cert) and a WebSocket route for Odoo's live bus (port 8072).

This is a small, self-contained follow-up to gaz 194. It only adds the ingress and its wiring; the `Service` and `Deployment` themselves belong to gaz 194.

## Ticket
GAZ-201 (sub-task of gaz 192, OpenSPP integration). Follow-up to gaz 194 (PR #157), which is now merged into `dev`, so this PR targets `dev` directly. The diff is only the ingress and its wiring; it edits files gaz 194 introduced (`openspp.sh`, `values.yaml`) and adds a template to the chart.

## What is included
- **Ingress template** `src/deployer/helm/openspp/templates/ingress.yaml`: gated on `ingress.enabled`; two paths: `/websocket` -> the `longpolling` port (8072), everything else `/` -> `http` (8069); TLS with the `openspp-tls` secret; NGINX annotations for long timeouts (3600s, for Odoo's long-poll), a 50m body size and SSL redirect.
- **Chart values** `src/deployer/helm/openspp/values.yaml`: a new `ingress` block (disabled by default, `className: nginx`, `host: ""`, `tls.secretName: openspp-tls`). The host is left empty on purpose so there is no hardcoded domain in the chart; `openspp.sh` sets it at deploy time.
- **Install script** `src/deployer/openspp.sh`: at deploy time it creates the self-signed `openspp-tls` secret with `create_ingress_secret` (the same helper `paymenthub.sh` use), and turns the ingress on via `--set ingress.enabled=true`, `ingress.host=<host>`, `ingress.tls.secretName=openspp-tls`, and `odoo.env.PROXY_MODE=true` (Odoo must know it runs behind a proxy so it trusts the `X-Forwarded-*` headers).
- **Host entry** `src/environmentSetup/environmentSetup.sh`: adds `openspp.$DOMAIN` to the `/etc/hosts` list alongside the other Gazelle hosts, so the browser resolves it locally.
- **Config** `config/config.ini`: a new `OPENSPP_DOMAIN` key in `[openspp]`. Empty by default, which means "derive `openspp.<GAZELLE_DOMAIN>`", so the domain still lives in one place. Set a full host to override.

## Out of scope (separate PRs)
- The OpenSPP deploy itself (chart, install script, wiring): gaz 194.
- Docs, smoke test and CI job: gaz 204.

### 1. Runtime request flow
```mermaid
flowchart LR
  B["Browser<br/>https://openspp.mifos.gazelle.test"]
  N["ingress-nginx controller<br/>terminates TLS (openspp-tls)"]
  ING["Ingress openspp<br/>host + 2 paths"]
  WS["/websocket -> longpolling 8072"]
  HTTP["/ -> http 8069"]
  SVC["Service openspp-odoo<br/>ClusterIP"]
  ODOO["Odoo pod<br/>PROXY_MODE=true"]

  B -->|"HTTPS"| N --> ING
  ING --> WS
  ING --> HTTP
  WS --> SVC
  HTTP --> SVC
  SVC --> ODOO
```
NGINX terminates TLS and forwards to the Odoo Service. Live-bus traffic (`/websocket`) goes to the longpolling port; everything else to the HTTP port. NGINX handles the WebSocket upgrade on its own; only the long timeouts are configured.

### 2. Deploy-time flow (what openspp.sh does for the ingress)
```mermaid
sequenceDiagram
  participant O as openspp.sh
  participant K as k8s cluster
  participant H as helm

  O->>O: resolve host = OPENSPP_DOMAIN or openspp.$GAZELLE_DOMAIN
  O->>K: create namespace openspp (if missing)
  O->>K: create_ingress_secret -> openspp-tls (self-signed, SANs incl. host + *.DOMAIN)
  O->>H: helm upgrade --install --set ingress.enabled=true, ingress.host=<host>,<br/>ingress.tls.secretName=openspp-tls, odoo.env.PROXY_MODE=true
  H->>K: create Ingress openspp (+ the gaz 194 objects)
```
The secret is created before helm runs, so the namespace must exist first (helm's `--create-namespace` happens later).

## How to test
OpenSPP uses a build-only image, so build/import it first exactly as in gaz 194 (`docker build ... ghcr.io/openmf/openspp:19.0` then `docker save ... | sudo k3s ctr images import -`). Then, from the repo root:

```bash
# Deploy (turns the ingress on automatically).
./run.sh -m deploy -a openspp

# The ingress object, with both paths and TLS.
kubectl get ingress -n openspp
kubectl describe ingress openspp -n openspp

# Reach the UI over HTTPS through the ingress (the acceptance criterion).
curl -k -o /dev/null -w 'HTTP %{http_code}\n' https://openspp.mifos.gazelle.test/web/login

# Remove it.
```

The host `openspp.<GAZELLE_DOMAIN>` resolves through `/etc/hosts`, where `setup-env` adds it. If it does not resolve on your machine, tell `curl` the IP with its `--resolve` option: read the node IP from `kubectl get nodes`, keep only the IPv4 address (a small regular expression that matches digits and dots does this), and pass that IP to `curl`.

## Evidence
Verified live on a full Gazelle stack (OpenSPP deployed, reached through the ingress over HTTPS):

```console
$ kubectl get pods -n openspp
NAME                                 READY   STATUS    RESTARTS   AGE
openspp-jobworker-56c85867cf-hfd2b   1/1     Running   0          19h
openspp-odoo-799d8979dd-2lrlb        1/1     Running   0          19h
openspp-postgis-0                    1/1     Running   0          19h

$ kubectl get ingress -n openspp
NAME      CLASS   HOSTS                        ADDRESS    PORTS     AGE
openspp   nginx   openspp.mifos.gazelle.test   10.0.1.4   80, 443   19h

# rules: /websocket -> openspp-odoo:longpolling (8072)
#        /          -> openspp-odoo:http (8069)
# TLS:   secret openspp-tls (kubernetes.io/tls), host openspp.mifos.gazelle.test

$ curl -sk -o /dev/null -w 'HTTP %{http_code}\n' https://openspp.mifos.gazelle.test/web/login
HTTP 200
$ curl -sk -o /dev/null -w 'HTTP %{http_code}\n' https://openspp.mifos.gazelle.test/web/health
HTTP 200
$ curl -s  -o /dev/null -w 'HTTP %{http_code}\n' http://openspp.mifos.gazelle.test/web/login
HTTP 308   # plain HTTP redirects to HTTPS (ssl-redirect)
```
Login and health both return 200 through the ingress over HTTPS, and plain HTTP redirects (308). Acceptance criterion met.

## Notes for reviewers
- **`host: ""` in values, set at deploy time**: the chart never hardcodes a domain. `openspp.sh` resolves the host from `OPENSPP_DOMAIN` (config.ini), falling back to `openspp.${GAZELLE_DOMAIN}`.
- **`PROXY_MODE=true`**: without it Odoo builds absolute URLs and cookies from its own address, not the public host, which breaks logins behind a proxy.
- **Self-signed cert**: `create_ingress_secret` makes a local self-signed cert (no trusted CA), so browsers show a one-time warning. This is the same pattern the other Gazelle DPGs use.
- **WebSocket**: only the two long-timeout annotations are needed; ingress-nginx upgrades the connection automatically, so no `proxy-set-header Upgrade` config is required.

## Checklist
- [x] Safe to re-run: `helm upgrade --install`; the ingress is just another chart object.
- [x] No hardcoded domain in the chart (`host: ""`); the host comes from config / GAZELLE_DOMAIN.
- [x] `helm lint` clean; `helm template` renders 1 Ingress when enabled, 0 when disabled.
- [x] `bash -n` passes on `openspp.sh` and `environmentSetup.sh`.
- [x] `-a all` is unchanged (OpenSPP is still off by default; the ingress only renders when deployed).

## AI usage


config/config.ini
```diff
@@ -168,3 +168,6 @@ OPENSPP_IMAGE_TAG = 19.0
 # password) which also keep redeploys idempotent. Set a file path for prod/remote real secrets.
 OPENSPP_DB_PASSWORD_FILE =
 OPENSPP_ADMIN_PASSWORD_FILE =
+# Ingress host for the OpenSPP web UI (GAZ-201). Leave EMPTY to derive it from GAZELLE_DOMAIN
+# (openspp.<GAZELLE_DOMAIN>), so the domain lives in one place. Set a full host to override.
+OPENSPP_DOMAIN =
```
tdaly61: Do we ever want a separate OPENSPP_DOMAIN ?
I mean if we are deploying from Gazelle , even if later on into multiple node clusters I can't see why we would need a new domain name ??   don't take that as opinion ... you have likely thought this through more than me so let me know what you think

src/deployer/helm/openspp/templates/ingress.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.ingress.enabled }}
+# Exposes Odoo via the shared NGINX ingress. Two paths: /websocket goes to the longpolling
+# port (8072, Odoo's live bus), everything else to HTTP (8069). ingress-nginx handles the
+# WebSocket upgrade automatically; only the long timeouts below are needed.
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "openspp.fullname" . }}
```
tdaly61, line 8: Soo if we don't need a separate DOMAIN NAME other than GAZELLE_DOMAIN , does this affect the ingress at all ??

config/config.ini
Ismael-Sallami: I added it for clarity, in case anyone wanted a different domain, but you are right and I
completely agree. Gazelle deploys everything under GAZELLE_DOMAIN, and no other component has
its own domain key, so OpenSPP should not be the exception. I have dropped it.

src/deployer/helm/openspp/templates/ingress.yaml
Ismael-Sallami: No, nothing changes here. The template only reads .Values.ingress.host, and the deployer is what
computes that value, so the host is now always openspp.${GAZELLE_DOMAIN} and the ingress stays as it is.

tdaly61 (approved):

PR #161, opened 2026-07-23
Fix vnext point transfers-command-handler-svc SCHEDULING_SVC_URL to scheduling-api-svc:3150

### LINKED ISSUE #160: Transfers-command-handler-svc points SCHEDULING_SVC_URL to a non-existent service and port
## Summary
transfers-command-handler-svc cannot reach the scheduling service. Its `SCHEDULING_SVC_URL` points at
`scheduling-svc:1234`, but the service deployed in the `vnext` namespace is `scheduling-api-svc` on port
`3150`. Every reminder fails with a DNS resolution error.

## Root cause
Wrong value (consumer), `src/deployer/manifests/vnext/apps/transfers-command-handler-svc-deployment.yaml`:

```yaml
- name: SCHEDULING_SVC_URL
  value: http://scheduling-svc:1234
```

Actual service, `src/deployer/manifests/vnext/crosscut/scheduling-api-svc-service.yaml` (`name: scheduling-api-svc`):

```bash
$ kubectl get svc -n vnext | grep -i sched
scheduling-api-svc   ClusterIP   10.43.31.116   3150/TCP
```

Both the hostname and the port are wrong.

## Evidence
Consumer logs, repeated on every reminder:

```text
(transfers-bc command-handler-svc v0.5.1) (SchedulingAdapter) error: Error: fetch failed
  cause: Error: getaddrinfo ENOTFOUND scheduling-svc  code: 'ENOTFOUND'  hostname: 'scheduling-svc'
```

In-cluster DNS:

```bash
$ kubectl exec -n vnext deploy/transfers-command-handler-svc -- getent hosts scheduling-svc
command terminated with exit code 2        # does not resolve
$ kubectl exec -n vnext deploy/transfers-command-handler-svc -- getent hosts scheduling-api-svc
10.43.31.116  scheduling-api-svc.vnext.svc.cluster.local  ...   # resolves
```

## Fix
```yaml
- name: SCHEDULING_SVC_URL
  value: http://scheduling-api-svc:3150
```

## Impact
Non-fatal. Reminders and timers (transfer retries) fail. End-to-end payments still complete because they
are currently sent serially, which does not rely on the scheduler. Worth fixing for correctness and to
remove the recurring error.

## Environment
mifos-gazelle dev branch, vnext, transfers-bc command-handler-svc v0.5.1, single-node k3s.

Ismael-Sallami:
## What
transfers-command-handler-svc could not reach the scheduling service. SCHEDULING_SVC_URL pointed at
scheduling-svc:1234, but the deployed service is scheduling-api-svc on port 3150, so every reminder
failed with `getaddrinfo ENOTFOUND scheduling-svc`.

## Fix
One value in one file:
  src/deployer/manifests/vnext/apps/transfers-command-handler-svc-deployment.yaml
  - value: http://scheduling-svc:1234
  + value: http://scheduling-api-svc:3150

## Evidence
```yaml
$ kubectl get svc -n vnext | grep -i sched
  scheduling-api-svc  ClusterIP  10.43.31.116  3150/TCP
$ kubectl exec -n vnext deploy/transfers-command-handler-svc -- getent hosts scheduling-svc      # fails
$ kubectl exec -n vnext deploy/transfers-command-handler-svc -- getent hosts scheduling-api-svc  # resolves
``` 

## Impact
Non-fatal: reminders and timers (transfer retries) fail. End-to-end payments still complete because
they are sent serially, which does not rely on the scheduler.

Closes #160


src/deployer/manifests/vnext/apps/transfers-command-handler-svc-deployment.yaml
tdaly61, line 1: Hi @Ismael-Sallami , this is likely fine and needed BUT I am confused .. why is this working today. Is the code path somehow not used except by OpenSPP ?

Ismael-Sallami: Hi Tom, it is not an OpenSPP-only path. The OpenSPP connector only talks to Payment Hub, and Payment Hub reaches the switch through connector-mojaloop, so every transfer hits the same command handler.

It works today because this URL is only used by the SchedulingAdapter for reminders/timers, not for prepare or fulfil. The call fails, the adapter logs it, and the transfer still commits and books. I saw the error on every transfer while those same transfers were completing, so the symptom is log spam plus a timeout path that never arms.

The wrong value was already in the vendored manifest, copied from upstream: their k8s manifest has `scheduling-svc:1234`, while their own compose `.env.sample` has the correct `scheduling-api-svc:3150`. I spotted it in the logs while debugging, and the fix was one value, so I opened the issue and this PR.

tdaly61: Ahh ..ok thanks !

tdaly61 (approved):

PR #166, opened 2026-07-28
GAZ-317: Build images for the host architecture and support arm64

Ismael-Sallami:
## Summary
Gazelle can import a locally built image into k3s, but it cannot build one. Any image that is not published to a registry has to be built by hand, on the right machine and for the right architecture.

This PR adds the missing half: one shared utility that builds an image with `docker buildx` for the **detected host architecture** and then imports it, reusing the existing import script rather than replacing it. It also lets the OpenSPP deployer pass the architecture and the database image to the chart, which is what makes an arm64 deployment possible. Building the OpenSPP image on an arm64 machine surfaced a bug in that import script which had to be fixed for the reuse to work at all.

## Ticket
GAZ-317. Cut from current `dev`.

## What is included
- **`src/utils/build-and-import-image.sh`** (new): builds with `docker buildx` for the host architecture (`uname -m`), then calls `src/utils/import-local-image-to-k3s.sh` to load the image into k3s, so there is one implementation of the import and one new one of the build. It takes `--context`, `--dockerfile`, `--target` and `--build-arg`, so it is tied to no build system: it serves the Java images and the Odoo one alike.
- **`src/utils/import-local-image-to-k3s.sh`**: two fixes, so the import works for registry-qualified names and without a tty.
- **`src/deployer/openspp.sh`**: detects the node architecture and passes `arch` to the chart, plus `postgis.image.repository` and `postgis.image.tag`, all three overridable by environment variable. The defaults are the chart's current values, so amd64 behaviour does not change.
- **`docs/BUILDING-IMAGES.md`** (new) and one row in the README documentation table. It is a repo-level document because the utility is not OpenSPP specific.

## Out of scope
- **Building an arm64 PostGIS image.** The official `postgis/postgis` has no arm64 build. This PR only makes the database image overridable and documents a third-party build as an experimental option for a demo; the chart default is unchanged.
- **Publishing images to a registry.** The `--push` path exists for a future multi-architecture image, but nothing is published here.

### 1. Two output paths, and why they are separate
```mermaid
flowchart LR
    A["build-and-import-image.sh"] --> B{"one platform<br/>or several?"}
    B -->|"host arch (default)"| C["buildx --load"] --> D["docker save"] --> E["k3s ctr images import"]
    B -->|"--platform a,b + --push"| F["buildx --push"] --> G["registry"]
    B -->|"--platform a,b, no --push"| H["error: explains why"]
```
A multi-platform build produces a **manifest list**, and only a registry can store one: it cannot be exported with `docker save`, so it cannot be imported into k3s. Asking for several platforms without `--push` therefore fails with that explanation instead of producing an artifact the local path cannot consume.

### 2. The import bug this had to fix first
The import script built its temporary tar path from the image name:

```bash
tarfile="/tmp/${IMAGE_NAME}.tar"
```

A name that carries a registry or an organisation contains slashes, and those become directories:

```
==> export docker image using docker save --output /tmp/ghcr.io/openmf/openspp.tar ghcr.io/openmf/openspp
failed to save image: invalid output path: stat /tmp/ghcr.io/openmf: no such file or directory
ERROR: Failed to save Docker image
```

So it only ever worked for bare names such as `myapp`, which is what its own help examples use, and every image built locally has at least an organisation in its name. Fixed by flattening the slashes for the temporary file only:

```bash
tarfile="/tmp/$(echo "$IMAGE_NAME" | tr '/' '_').tar"
```

It belongs in this PR because the new utility delegates the import instead of duplicating it, and with this bug the delegation cannot work for any real image. The second fix, in the same function, resolves the invoking user from `$SUDO_USER` and keeps `who am i` as a fallback, with a loud error if neither yields anything: `who am i` reads the session record and comes back empty without a tty, so on its own it fails over a non-interactive ssh or in CI.

### 3. Why the pods used to stay Pending on arm64
The chart has always had an `arch` value that becomes a `nodeSelector` on the three workloads, but nothing set it, so it kept its `amd64` default. On an arm64 node the scheduler then finds no matching node and the pods sit `Pending` with no visible cause. The deployer now reads `uname -m`, passes it through, and says out loud what it decided:

```
INFO   Node architecture: arm64 (db image imresamu/postgis:18-3.6.1-alpine3.22)
```

## How to test

```bash
# Help, and the guard that refuses a multi-platform local build.
src/utils/build-and-import-image.sh --help
src/utils/build-and-import-image.sh -n foo -t 1 --platform linux/amd64,linux/arm64

# Chart rendering, no cluster needed: arm64 selector, no selector, unchanged db image.
helm template openspp src/deployer/helm/openspp --set arch=arm64 | grep -c "kubernetes.io/arch: arm64"   # 3
helm template openspp src/deployer/helm/openspp --set arch=      | grep -c nodeSelector                  # 0
helm template openspp src/deployer/helm/openspp | grep -E "image: .*postgis"    # postgis/postgis:18-3.6-alpine

# Build and deploy on arm64 hardware.
src/utils/build-and-import-image.sh -n ghcr.io/openmf/openspp -t 19.0 \
    -c <OpenSPP2 checkout> -f <OpenSPP2 checkout>/docker/Dockerfile --target production
OPENSPP_POSTGIS_REPOSITORY=imresamu/postgis OPENSPP_POSTGIS_TAG=18-3.6.1-alpine3.22 \
    ./run.sh -m deploy -a openspp

# Teardown.
```

## Evidence
Run on an OCI Ampere A1 instance (`VM.Standard.A1.Flex`, 4 OCPUs, `Neoverse-N1`), alongside an existing deployment. The image is 2.24 GB; the first build there, with an empty cache, took about 28 minutes, and all three pods were Ready in under two minutes.

```console
  platform   : linux/arm64 (host arch: arm64)
 ** built ghcr.io/openmf/openspp:19.0 (architecture: arm64)
==> export docker image using docker save --output /tmp/ghcr.io_openmf_openspp.tar ghcr.io/openmf/openspp
 ** images appear to have imported ok

$ sudo k3s ctr images ls | grep openspp:19.0
ghcr.io/openmf/openspp:19.0   2.2 GiB   linux/arm64

INFO   Node architecture: arm64 (db image imresamu/postgis:18-3.6.1-alpine3.22)   # with -d true

POD                                 READY   IMAGE                                  NODE
openspp-jobworker-8bcf94f4d-zxlm5   true    ghcr.io/openmf/openspp:19.0            ocifree3
openspp-odoo-5bd46b6dc4-gxgjx       true    ghcr.io/openmf/openspp:19.0            ocifree3
openspp-postgis-0                   true    imresamu/postgis:18-3.6.1-alpine3.22   ocifree3

nodeSelector: {"kubernetes.io/arch":"arm64"}      (all three)
GET /web/health -> 200
spp_base_common -> installed                      (53 modules installed)
SELECT postgis_version() -> 3.6 USE_GEOS=1 USE_PROJ=1 USE_STATS=1
```

Both the health check and the module check are needed: an empty Odoo also answers 200 on `/web/health`, so the HTTP code alone does not prove the module is there. The full run, command by command, is attached to the ticket.

## Notes for reviewers
- **The architecture mapping is not new.** It is the same `uname -m` case already used in `src/environmentSetup/k8s.sh`, so the repo keeps one naming convention.
- **The builder is pinned.** A single-platform build uses the built-in `default` builder, which writes straight into the local image store that `docker save` reads. Left unpinned, buildx uses whichever builder the machine happens to have selected, which is per-user configuration nobody looks at. `--builder` overrides the choice, because on Docker Desktop the docker-driver builder is not always called `default` and I had no macOS machine to test that on. The `docker-container` builder is created only for a real multi-platform build, which is the only case that needs it.
- **The import step is Linux-only, by nature.** It ends in `k3s ctr images import`, so it needs k3s on the same machine. On macOS the cluster runs inside Colima and the host has no `k3s` binary, so run the utility inside that machine or use `--push`. The build half works anywhere Docker does.
- **BuildKit is required**, not preferred: the OpenSPP2 Dockerfile reads `TARGETARCH` and uses cache mounts, and it fails on purpose when `TARGETARCH` is empty. `--load` matters for the same reason: without it a `docker-container` builder leaves nothing for `docker save` to find.
- **The database image override reaches three places.** Besides the PostGIS StatefulSet, that image is used for a `wait-for-postgres` initContainer in both the Odoo and the jobworker Deployments, because it carries `pg_isready`. On an arm64 node an amd64 initContainer would fail with `exec format error` and the pod would never reach its application container.
- **What is not exercised.** The `--push` path is implemented but has not been run: it needs registry credentials. Everything on the local path is covered by the evidence above.

## Checklist
- [x] Multi-platform without `--push` fails with an explanatory error.
- [x] Import delegated to the existing script, not duplicated, and working for a registry-qualified name.
- [x] Import works with no tty, over a non-interactive ssh, which is the CI case.
- [x] arm64: image builds on arm64 hardware and OpenSPP starts (3 pods Ready, health 200, base module installed).
- [x] amd64: deploy unchanged. The rendered manifests are byte-identical with and without the new `--set` values, and a full deploy was run on an amd64 cluster.
- [x] `values.yaml` untouched: the defaults come from the chart.
- [x] No new dependency beyond the buildx plugin, and the utility says how to install it.
- [x] Documented in `docs/` and linked from the README.

## AI usage


docs/BUILDING-IMAGES.md
```diff
@@ -0,0 +1,85 @@
...
+runtime. `src/utils/build-and-import-image.sh` does both, for the architecture of the host it runs on.
+
+## Quick start
+
+```bash
+# build for the host architecture and import into k3s
+src/utils/build-and-import-image.sh -n ghcr.io/openmf/openspp -t 19.0 \
+    -c <OpenSPP2 checkout> -f <OpenSPP2 checkout>/docker/Dockerfile --target production
+
+# build a multi-architecture image and push it to a registry
```
tdaly-mifos, line 14: How does push to a registry work ?? Where do the registry credentials come from ?

src/utils/build-and-import-image.sh
```diff
@@ -0,0 +1,208 @@
...
+    local arch
+    arch=$(uname -m)
+    case "$arch" in
+        x86_64)       echo "amd64" ;;
+        aarch64|arm64) echo "arm64" ;;
+        *) error "Unsupported architecture: $arch (only amd64 and arm64 are supported)" ;;
+    esac
+}
+
+while [[ $# -gt 0 ]]; do
```
tdaly-mifos, line 90: Yeah same question, I am not understanding "how" the registry is identified and how to enable authentication and authorisation

src/utils/import-local-image-to-k3s.sh
```diff
@@ -44,7 +44,11 @@ error() {
 
 set_user() {
```
tdaly-mifos, line 45: Oh I wonder if any of the docker build now needs to be run as sudo ?  It might but can you verify please , because I had vastly reduced the use of SUDO but I think I have plenty of "left overs" in  the src/utils .sh scripts

src/utils/build-and-import-image.sh
tdaly-mifos, line 1: but does this script import ?
I am just wondering if there is a simplification for the code and the user by combining the 2 scripts or if they should be separate as you have them

tdaly-mifos:
@Ismael-Sallami you say 
A multi-platform build produces a manifest list, and only a registry can store one: it cannot be exported with docker save, so it cannot be imported into k3s. Asking for several platforms without --push therefore fails with that explanation instead of producing an artifact the local path cannot consume.

really ?  Ok I am not arguing just seeking confirmation 
Can you please look at this script https://github.com/mojaloop/platform-shared-tools/blob/main/packages/installer/mini-loop/ttk-interim-fix.sh what is did was clone the repo, build the component and import to k3s ?    This is mainly what we are after for your script , so we can run ./run.sh -m deploy -a openspp and arm or amd ..it just works

tdaly-mifos (commented):
hi @Ismael-Sallami , can you please tell me if you have tested this on ARM Ubuntu such that run.sh -a openspp deploys cleanly and your demo runs ?

tdaly-mifos:
oh and @Ismael-Sallami , yeah your point about this not working on MacOS is a good one.  So I am torn on this one, I am happy to fix this later as Mac is NOT the primary platform BUT I do want you to have the owenership of the openspp project.  So lets move ahead with ubuntu and discuss MacOS later

Ismael-Sallami:
> @Ismael-Sallami you say A multi-platform build produces a manifest list, and only a registry can store one: it cannot be exported with docker save, so it cannot be imported into k3s. Asking for several platforms without --push therefore fails with that explanation instead of producing an artifact the local path cannot consume.
> 
> really ? Ok I am not arguing just seeking confirmation Can you please look at this script https://github.com/mojaloop/platform-shared-tools/blob/main/packages/installer/mini-loop/ttk-interim-fix.sh what is did was clone the repo, build the component and import to k3s ? This is mainly what we are after for your script , so we can run ./run.sh -m deploy -a openspp and arm or amd ..it just works

---

Good question Tom, I had the same doubt, and I think I was too absolute there.

The behaviour I described is real: with Docker's **classic** image store the build fails with `ERROR: failed to build: docker exporter does not currently support exporting manifest lists`. I get that with both the `docker` and the `docker-container` driver, so the limit is the destination, not the builder. But Docker Engine 29 makes the **containerd image store** the default on fresh installs, and that store does hold manifest lists. My daemon came from upgrades, so it kept the classic overlay2 store, which is what upgraded installs keep and where that error comes from. So I will rewrite that paragraph.

I also tested the part I got wrong. With `--output type=oci` a two-platform build goes into a plain tarball, and `sudo k3s ctr images import --all-platforms` loaded it into my k3s as `application/vnd.oci.image.index.v1+json`. So a manifest list does **not** need a registry.

The behaviour of the script still stays, for two better reasons. `ctr images import` imports content for a single platform unless you pass `--all-platforms` (its own help says "false by default"), so a multi-arch import into one node pays for two builds and uses one. And building the other architecture on the same machine goes through QEMU, which the Docker docs say "can be much slower than native builds". Our Odoo image already takes ~28.5 minutes natively on the ARM box, so I did not try the emulated route.

On your script: the three middle steps are the same as mine, `docker build` with no `--platform`, `docker save`, `k3s ctr images import`. The two ends are not: I do not clone (the utility takes an existing checkout as the build context) and I do not rewrite manifests, because `openspp.sh` passes the image to the chart with `--set odoo.image.repository/tag`.

And you are right that today it does not "just work": `openspp.sh` checks whether the image is in the node and fails with instructions. My plan is `OPENSPP_BUILD_IF_MISSING` in `[openspp]`: if the image is missing and the flag is on, fetch OpenSPP2 at its pinned tag and call the same utility, then deploy. With the flag off it behaves exactly as today. Two things to flag: `run.sh` refuses to run as root while the import needs it, so the `sudo` stays inside and only around the import; and a cold build is around half an hour, so the deploy must say that before it starts. 

**One question on process**: this PR is already validated on arm64 and amd64, so I would rather land the wording fixes here and do `OPENSPP_BUILD_IF_MISSING` as its own small PR with its own sub-task, which also keeps it out of the way of #158 touching the same file. Happy to put it all in here if you prefer one round.

tdaly-mifos:
hi @Ismael-Sallami , yeah as this doesn't touch too much I am happy to approve and merge then apply fixes later.

tdaly61 (approved):
see comments from tdaly-mifos

Ismael-Sallami:
Thanks for merging, and for the quick turnaround.

I spent this morning on your ARM question, so here is the rest in one message.

**ARM Ubuntu: yes, both halves.** Devarsh told me I could clear the OCI box, so I deployed the full stack there. `run.sh -a openspp` comes up clean, and the demo runs end to end: **5 of 5 subsidies credited in Fineract, verification PASS, first try**. All five DPGs fit.  One extra on resources: vnext and mifosx declare no CPU requests, so only infra and paymenthub reserve anything, and it still all ran fine on those four cores once I scaled six interfaces to 0.

**Registry and credentials.** The utility does not handle authentication at all: there is no login, token or credential handling in it. The registry comes from the image name, so `ghcr.io/openmf/openspp` goes to ghcr.io and a bare name goes to Docker Hub. What you need is a `docker login` to the target registry first; the Docker CLI stores that in `~/.docker/config.json`, or in a credential helper if one is configured. I have no login on this machine, so the push fails with a 403 when it asks ghcr.io for an anonymous token. Worth separating from `config.ini [dockerhub]`: the comment there says those are for authenticated image pulls, and the script that uses them creates a Kubernetes pull secret with the server fixed to docker.io, so they never log the host in and do not help a push. Which registry and account we publish to, and the token it needs, I would rather agree with you than assume.

**Does the build need sudo: no.** The build runs as the normal user. The only `sudo` wraps the import, because `k3s ctr` talks to a root-owned socket. It works without sudo because Gazelle already adds the user to the `docker` group during environment setup.

**Does it import: yes**, that is the default, and `--no-import` turns it off. It does not duplicate `import-local-image-to-k3s.sh`, it calls it, so the `docker save` plus `k3s ctr images import` logic lives in one place only.

**On merging the two, I would keep them separate.** Building and importing do not always travel together: an image that arrives as a tar, or one built on another machine, needs importing and never needed building, and a `--push` build imports nothing. As one file it would need two modes and more flags to choose between them. And while looking at the import question I noticed src/utils/test-scripts/publish-local-image.sh is quite similar to import-local-image-to-k3s.sh, tar path included, in case that directory ever gets a tidy up.

**macOS**: agreed, Ubuntu first. The reason it cannot work today is that the import needs k3s on the same machine, and on macOS the cluster lives inside Colima. The fix later is to run the import through `colima exec`, which is the pattern the environment setup already uses there.

I will raise the PR with those wording changes. I hope this answers everything; if not, please do ask.

PR #167, opened 2026-08-02
GAZ-307: Deploy Credit Bureau with initContainer dependency on Fineract

kanishksingh23:
## Summary
Deploys the Mifos X Credit Bureau plugin into the MifosX stack. It shares the existing infra Postgres via a dedicated `creditbureau` database. One initContainer gates Credit Bureau startup until Fineract's readiness endpoint is green; a second, idempotent initContainer ensures the `creditbureau` database exists (so it deploys cleanly on both fresh and existing clusters).

Adds two manifests (`credit-bureau-deployment.yaml`, `credit-bureau-service.yaml`) plus the `creditbureau` database in the infra Postgres init. No `mifosx.sh` change needed, the manifests dir is applied automatically.

## Note
- Image `kanishksingh23/mifos-credit-bureau:30072026` is a temporary multi-arch (amd64 + arm64) build from source with the Postgres driver added (openMF/mifos-x-credit-bureau-plugin#141), to be republished under `openMF` and re-pinned.
- Verified via `./run.sh -m deploy -a mifosx`: Credit Bureau holds at `Init` until Fineract is Ready, then reaches `1/1 Running`; `GET /credit-bureaus` returns 200 through the Service.

### DIFF
--- a/src/deployer/helm/infra/values.yaml
+++ b/src/deployer/helm/infra/values.yaml
@@ -351,4 +351,5 @@ postgresql:
         01-create-fineract-dbs.sql: |
           CREATE DATABASE fineract_tenants;
           CREATE DATABASE fineract_default;
+          CREATE DATABASE creditbureau;
 
--- a/src/deployer/manifests/mifosx/credit-bureau-deployment.yaml
+++ b/src/deployer/manifests/mifosx/credit-bureau-deployment.yaml
@@ -1,0 +1,106 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: credit-bureau
+    tier: backend
+  name: credit-bureau
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: credit-bureau
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: credit-bureau
+        tier: backend
+    spec:
+      initContainers:
+        - name: ensure-creditbureau-db
+          image: postgres:16-alpine
+          env:
+            - name: PGPASSWORD
+              value: "postgrespw"
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -e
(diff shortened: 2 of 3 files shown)


tdaly61 (commented):
hi @kanishksingh23 hey does this PR require liquibase = true ?  reason for asking is that I am not understanding how this works without it and it it does require it then it will take a long time and I would think that maybe doing this load and then creating a new DB dump would be faster at deploy time.  

Also have you done any readme updates for this ??

tdaly61:
Oh also can you please look at the merge conflicts please @kanishksingh23

kanishksingh23:
Hey @tdaly61, thanks for the review.

On liquibase: no, this doesn't need liquibase = true. The plugin actually ships both a liquibase changelog (src/main/resources/db/changelog/) and Hibernate ddl-auto, but by default liquibase is disabled (spring.liquibase.enabled: false) and Hibernate ddl-auto: update builds the schema at boot. The deployment just uses that default, it sets no liquibase flag.

Either way it's a non-issue for deploy time: the CB schema is only 3 tables (credit_bureau, cb_registration_params, credit_bureau_registration_param_values), so whichever mechanism builds it, it's effectively instant, nothing like Fineract's changelog, so no load-then-dump step is needed.

I verified it via ./run.sh -m deploy -a mifosx: the ensure-creditbureau-db initContainer creates the empty DB, CB boots, and the pod logs show Hibernate initialising the schema with no Liquibase activity at all. The only wait in CB's startup is the wait-for-fineract initContainer (blocking on Fineract readiness, which has to be up regardless); then tables are created, GET /credit-bureaus returns 200, and the pod reaches 1/1.

About the readme updates, I am thinking of doing this in the docs ticket that we discussed earlier.

kanishksingh23:
> Oh also can you please look at the merge conflicts please @kanishksingh23

I have resolved the merge conflicts.

tdaly61 (approved):

PR #162, opened 2026-07-25
GAZ-204 Integrate into Gazelle v1.4.0

Ismael-Sallami:
## Summary
OpenSPP (OpenSPP2 / Odoo 19) can be deployed by Gazelle (`./run.sh -m deploy -a openspp`, from gaz 194) and reached over HTTPS (from gaz 201). This PR is the last integration step: it makes OpenSPP a first-class, documented and tested part of Gazelle. It adds a deploy guide, a smoke test that proves a deploy is really healthy (not just returning 200 from an empty Odoo), and a CI job that builds the image, deploys OpenSPP and runs that smoke test.

No product code changes. Only docs, a test, a CI job, and one row in the main `README.md`.

## Ticket
GAZ-204 (sub-task of gaz 192, OpenSPP integration). Cut off current `dev`, which already includes the OpenSPP deploy (gaz 194, PR #157). It does not depend on the ingress PR (gaz 201, #158): the smoke test checks `/web/health` from inside the odoo pod, so no ingress or hosts entry is needed.

## What is included
- **Deploy guide** `docs/OPENSPP.md`: what OpenSPP is and why it fits Gazelle (it is the social registry the other DPGs lack), what it is in Gazelle (optional DPG, off by default, own PostgreSQL 18 + PostGIS 3.6), how to build/import the build-only image, the `[openspp]` config keys, deploy/access/teardown, what to do once it is up (the Registry / Apps / Settings menus), and the amd64-only limitation.
- **README entry** `README.md`: one row in the documentation table and one quick-start line, in the same place and style as the other DPGs, so the doc is discoverable where a new user looks first.
- **Smoke test** `tests/openspp/smoke.sh`: three checks against a live deploy: pods Ready, the base module `spp_base_common` is **actually installed** (queried in PostGIS, because a bare Odoo also returns 200), and `/web/health` returns 200. Namespace overridable with `OPENSPP_NAMESPACE`.
- **CI job** `.circleci/config.yml`: a new `deploy_and_verify_openspp` command and a `deploy-test-openspp-amd64` job wired into the `test-multi-arch` workflow. It sets up k3s, builds and imports the build-only OpenSPP2 image, deploys OpenSPP, and runs the smoke test. amd64 only.

## Out of scope (separate PRs)
- The OpenSPP deploy itself (chart, install script, wiring): gaz 194.
- The NGINX ingress that exposes OpenSPP over HTTPS: gaz 201.

### 1. CI job flow
```mermaid
sequenceDiagram
[...]

PR #170, opened 2026-08-04
GAZ-321: build the OpenSPP image during the deploy when it is missing

Ismael-Sallami:
## Summary
`./run.sh -m deploy -a openspp` now works on a clean machine without a manual build step. When the OpenSPP2 image is neither in the cluster nor pullable, the deploy fetches the source, builds the image for the host architecture, imports it and carries on. A switch in `config.ini` turns that off, and with it off the deploy behaves exactly as it does today.

The same script gets two smaller fixes that belong to the same experience: the smoke test no longer reports a correct deploy as failed, and the closing message names the ingress URL like the other DPGs do.

## Ticket
GAZ-321 (sub-task of gaz 192, OpenSPP integration). Cut off current `dev`, which already has the shared build utility (gaz 317, PR #166) and the smoke test this now delegates to (gaz 204, PR #162).

Follow-up to the review on gaz 317, where the ask was that the deploy just works on a clean machine, on amd64 or arm64.

## What is included
- **`src/deployer/openspp.sh`**: the three-state decision, the source fetch, the build call, the smoke test delegated, and the closing message.
- **`src/utils/build-and-import-image.sh`**: five lines, so the utility refuses early when the import cannot work.
- **`config/config.ini`**: the switch and the two source keys.
- **`docs/OPENSPP.md`**: prerequisites rewritten for the new behaviour.

## Out of scope
- Publishing the image anywhere. That is the open question about image hosting, and this PR does not answer it: it makes the local build painless and gets out of the way the day an image is published.
- Bumping the pinned OpenSPP2 release. The validated image comes from the pinned tag, so moving it means a rebuild and a full revalidation.

### 1. Three states, cheapest first
```mermaid
flowchart TD
  A["image already in the cluster?"] -->|yes| D["deploy, nothing else to do"]
  A -->|no| B["can it be pulled from its registry?"]
  B -->|yes| E["deploy, Kubernetes pulls it"]
  B -->|no| C["build enabled, and can it be imported here?"]
  C -->|yes| F["fetch source, build, import, deploy"]
  C -->|no| G["stop, saying which way out applies"]
```

Reading the node is free, asking the registry is one short request, building is about half an hour. So the usual case, where the image is already there, makes no network call at all.

The registry check is the part that matters beyond today. A published image is not on the node either on a clean machine: the kubelet pulls it when the pod starts, because the chart uses `IfNotPresent`. So the deploy asks whether the image can be pulled and only builds when it cannot. Pointing at a published image later is two keys in `config.ini` and no code change.

### 2. Why the import cannot always happen, and why that is checked first
An image lives in the container runtime of each node, and there is no API to upload one to a cluster. The import runs `k3s ctr` against the containerd socket **of the machine it runs on**, so it only works where that machine is the node. It does not work against a remote cluster, and it does not work on macOS, where the cluster lives inside the Colima VM.

Before this change the failure came **after** the build: half an hour of work and then `k3s: command not found`. Now the deploy checks first and says which way out applies, either pushing the image to a registry or building it on the node itself. The same check went into the build utility next to its docker and buildx checks, so anything else that uses it is covered too.

### 3. The smoke test
It opened a port-forward on a fixed local port, waited five seconds and made a single request. When that port was already taken the forward could not bind, the error went to `/dev/null`, and the request came back with no HTTP code, so a **correct** deploy was reported as failed. It now runs `tests/openspp/smoke.sh`, which checks the pods, the installed module and `/web/health` through the cluster API and opens no local port.

## How to test
```bash
# The image is already there: nothing is built.
./run.sh -m deploy -a openspp -d true

# Switch off with no image: the message of today plus the line naming the switch.
sudo k3s ctr images rm ghcr.io/openmf/openspp:19.0
OPENSPP_BUILD_IF_MISSING=false ./run.sh -m deploy -a openspp

# A pullable image: nothing is built, the deploy proceeds to the pull.
OPENSPP_IMAGE_REPOSITORY=postgis/postgis OPENSPP_IMAGE_TAG=18-3.6-alpine \
    ./run.sh -m deploy -a openspp -d true

# Switch on with no image: builds, imports and the pods come up Ready.
sudo k3s ctr images rm ghcr.io/openmf/openspp:19.0
free -h && df -h /          # a cold build is about 30 minutes and 3 GB
./run.sh -m deploy -a openspp -d true
```

The pullable case does not bring the pods up, because that image is not OpenSPP. It tests the branch that is taken, which is visible in the log, and not the deployment.

## Evidence
Run on an amd64 machine with the whole stack up, and on an arm64 machine. The buildx output is long and repetitive, so only its header, its total and the import time are kept.

Image already there, so nothing is built:

```console
$ ./run.sh -m deploy -a openspp -d true
==> Deploying OpenSPP
INFO   Using image ghcr.io/openmf/openspp:19.0
INFO   Node architecture: amd64 (db image postgis/postgis:18-3.6-alpine)
    Waiting for PostGIS and Odoo to become Ready            [  ok  ]
    Smoke test (/web/login + OpenSPP module)                INFO   OpenSPP2 smoke test (namespace: openspp)
  PASS  openspp-postgis pod is Ready
  PASS  openspp-odoo pod is Ready
  PASS  module spp_base_common is installed
  PASS  /web/health returns 200

OpenSPP2 smoke test: OK

  OpenSPP:   https://openspp.mifos.gazelle.test/web/login   (admin / admin)
```

Image removed and the build switched off. It stops before the TLS secret and before Helm:

```console
$ OPENSPP_BUILD_IF_MISSING=false ./run.sh -m deploy -a openspp
==> Deploying OpenSPP
ERROR  Image ghcr.io/openmf/openspp:19.0 is not in the cluster and cannot be pulled.
ERROR  Build and import it with:  src/utils/build-and-import-image.sh -n ghcr.io/openmf/openspp -t 19.0 ...
ERROR  Or import an image you already built:  docker save ghcr.io/openmf/openspp:19.0 | sudo k3s ctr images import -
ERROR  Set OPENSPP_BUILD_IF_MISSING = true in config.ini [openspp] to build it automatically.
```

Switched on, from no image and no source:

```console
$ ./run.sh -m deploy -a openspp -d true
INFO   Image ghcr.io/openmf/openspp:19.0 is missing and OPENSPP_BUILD_IF_MISSING is true.
INFO   Building it now. A cold build takes about 30 minutes and around 3 GB of disk.
INFO   The import step runs sudo, so it may ask for your password.
INFO   Cloning OpenSPP2 v19.0.2.0.0 into .../repos/OpenSPP2

  platform   : linux/amd64 (host arch: amd64)
[+] Building 3.8s (28/28) FINISHED
Importing    elapsed: 16.2s
INFO   Image ghcr.io/openmf/openspp:19.0 built and imported.
    Waiting for PostGIS and Odoo to become Ready            [  ok  ]
  PASS  openspp-postgis pod is Ready
  PASS  openspp-odoo pod is Ready
  PASS  module spp_base_common is installed
  PASS  /web/health returns 200
```

Same again with the source left in place, so it is reused instead of cloned:

```console
INFO   Using the OpenSPP2 checkout at .../repos/OpenSPP2
[+] Building 2.8s (28/28) FINISHED
INFO   Image ghcr.io/openmf/openspp:19.0 built and imported.
```

And the same three cases on arm64. The line that matters is the platform, chosen from the host with nothing passed by hand:

```console
$ export OPENSPP_POSTGIS_REPOSITORY=imresamu/postgis
$ export OPENSPP_POSTGIS_TAG=18-3.6.1-alpine3.22
$ ./run.sh -m deploy -a openspp -d true

INFO   Cloning OpenSPP2 v19.0.2.0.0 into .../repos/OpenSPP2
  platform   : linux/arm64 (host arch: arm64)
[+] Building 3.7s (30/30) FINISHED
Importing    elapsed: 40.9s
INFO   Image ghcr.io/openmf/openspp:19.0 built and imported.
INFO   Node architecture: arm64 (db image imresamu/postgis:18-3.6.1-alpine3.22)
  PASS  openspp-postgis pod is Ready
  PASS  openspp-odoo pod is Ready
  PASS  module spp_base_common is installed
  PASS  /web/health returns 200
```

The builds took seconds because the layer cache was warm on both machines. A cold build is the 30 minutes the deploy warns about. The database image is overridden on arm64 only, because the official PostGIS image publishes an amd64 manifest only.

## Notes for reviewers
- **With the switch off, nothing changes.** Same behaviour and same error message as today, plus one line naming the switch. That is what makes this reviewable.
- **The registry probe uses `docker manifest inspect`, which Docker still labels experimental.** It works without enabling anything and downloads no layer, only the manifest. `docker buildx imagetools inspect` is the stable alternative and returns the same exit codes, but it needs the buildx plugin, which the probe otherwise does not. If the probe cannot answer, the code falls through to building, which is a safe failure.
- **The probe and the kubelet do not share credentials.** The probe uses the host's docker configuration, the kubelet uses the node's. With a public registry that does not matter. With a private one the probe could say yes and the pod still end in `ImagePullBackOff`, and no Gazelle chart uses `imagePullSecrets` today.
- **A failed pull attempt cannot be told apart from a missing image.** A registry answers `denied` both for something that does not exist and for something private without login, so the message says the image cannot be pulled from here and does not claim it does not exist.
- **The pinned OpenSPP2 release stays as it is**, and it is visible in the configuration table of the documentation. Upstream has a newer release; moving to it means a rebuild and a revalidation, so it is not part of this PR.
- **The smoke test does not read the release name.** It uses the default resource names, while the code it replaces honoured `OPENSPP_RELEASE_NAME`. Nothing in Gazelle sets that key to anything else, so this is a note rather than a regression, and it is one line in the smoke test if it ever matters.
- **The build runs as the current user; only the import needs root**, and the utility handles that. The deploy says so before it starts.

## Checklist
- [x] With the switch off, behaviour and error message identical to today.
- [x] Nothing is built when the image is present or pullable.
- [x] Where the image cannot be imported, the deploy stops before building.
- [x] The build utility refuses early too, and only when an import is actually intended.
- [x] A clean deploy builds, imports and the pods come up Ready, on amd64 and on arm64.
- [x] The build targets the host architecture with nothing passed by hand.
- [x] An existing source checkout is reused instead of cloned again.
- [x] No new dependency: the utility, the smoke test and the config loader already exist.
- [x] `bash -n` clean on both scripts, and the configuration keys read back with `crudini`.

## AI usage


docs/OPENSPP.md
```diff
@@ -48,19 +48,28 @@ in `ODOO_INIT_MODULES`), so there is **no separate init Job**.
 ## Prerequisites
 
 1. A running cluster. On a fresh machine: `sudo ./setup-env.sh -u $USER` (installs k3s + tools).
-2. The OpenSPP2 image **loaded into the cluster**. It is build-only (no published image yet), so build
-   the Dockerfile `production` target from the OpenSPP2 source and import it into k3s:
+2. `docker` with the buildx plugin, because the OpenSPP2 image is not published anywhere yet and the
+   deploy builds it. See [Building images](BUILDING-IMAGES.md).
```
tdaly61: Yeah but this is not automated if docker buildx is not there then this breaks.  So we ought to fix that in the environmentSetup.sh ..please see the output plan from Claude that I am going to email you.

tdaly61 (changes requested):
@Ismael-Sallami I think just your doc has to change here as well and the changes to ensure that docker buildx is available i.e. your script should still test for docker buildx but we should make sure that test always passes by installing buildx.   The motivation here it "simplicity" for the end user.  Get things up , installed , running , get them to run a successful demo and then that becomes the point at which they can then start learning by probing things, asking questions, changing things etc

docs/OPENSPP.md
tdaly61: Actually thinking about this a bit more I am putting the plan below BUT I am not sure if we should do this or not as it depends on if we are going to host images and avoid having to do the local build in which case this is a one-off , still required by only required mostly by the gazelle team , and they are quite capable of installing docker buildx ..humm

in any case here is the claude plan ..but as I say hold off on this for the moment 
# Ensure `docker buildx` is available on both Ubuntu and macOS (Colima)

## Context

`src/utils/build-and-import-image.sh` already uses `docker buildx` to build (and optionally push multi-arch) images, but it only checks for buildx and errors out with an install hint (lines 111–113) rather than the environment ever installing it. `install_os_prerequisites` (Linux) and `install_k8s_tools`/`install_mac_k8s` (macOS) install Docker/Colima but never provision the buildx plugin, so a fresh machine following `setup-env.sh` will fail the first time it tries to build an image. The fix is to have buildx installed as part of the existing OS-prerequisite/tool-install flow on both platforms, matching how the rest of docker tooling is already provisioned.

## Changes

### 1. Ubuntu/Linux — `src/environmentSetup/environmentSetup.sh`, `install_os_prerequisites` (~line 32)

Docker's official apt repo is already configured just above this line (lines 29–31), so `docker-buildx-plugin` is available from the same repo. Add it to the existing install line:

```bash
apt install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin >> /dev/null 2>&1
```

Since this is inside the `if ! command -v docker &> /dev/null; then` block, also add a standalone check so buildx gets installed/repaired even when `docker` itself is already present (e.g. someone installed docker outside this script, or upgraded from an older gazelle run before this change):

```bash
if ! docker buildx version &> /dev/null; then
    log_with_verbose_check "$debug" debug "docker buildx is not installed. Installing docker-buildx-plugin..."
    apt-get update >> /dev/null 2>&1
    apt-get install -y docker-buildx-plugin >> /dev/null 2>&1
    log_ok
else
    log_with_verbose_check "$debug" debug "docker buildx is already installed.\n"
fi
```
Place this check right after the existing docker install if/else block (after line 37), alongside the nc/jq checks that already follow that pattern.

### 2. macOS — `src/environmentSetup/k8s.sh`, `install_k8s_tools` (line 200)

Add the `docker-buildx` Homebrew formula to the existing `mac_tools` array so it goes through the same idempotent install/unlink/relink loop as `docker`, `docker-compose`, etc.:

```bash
local mac_tools=(helm k9s kubectx kustomize kubectl docker docker-compose docker-buildx socket_vmnet)
```

Homebrew's `docker-buildx` formula installs the CLI plugin into `~/.docker/cli-plugins/docker-buildx`, not onto PATH as a standalone binary, so the loop's `command -v "$tool"` / `"$tool" version` checks won't work for it the same way as other tools. Adjust the loop (or add a small dedicated branch) to check with `docker buildx version &>/dev/null` instead of `command -v docker-buildx`, and just run `run_brew install docker-buildx` when that check fails — skip the unlink/relink steps for this one entry since it's not a linked binary. Simplest approach: pull `docker-buildx` out of the array and handle it as one extra explicit step after the loop:

```bash
if ! docker buildx version &>/dev/null; then
    run_brew install docker-buildx >/dev/null 2>&1
fi
```

### 3. Update the planned-changes summary text

`show_planned_changes` in `environmentSetup.sh`:
- Line 450 (apt list): `"apt: docker-ce jq netcat..."` → `"apt: docker-ce docker-buildx-plugin jq netcat..."`
- Line 444 (Homebrew list): `"Homebrew: colima docker helm k9s kubectx..."` → `"Homebrew: colima docker docker-buildx helm k9s kubectx..."`

### 4. Docs

`docs/BUILDING-IMAGES.md` lines 43–45 currently say buildx must be manually installed if missing. Update this note to reflect that `setup-env.sh` now provisions buildx automatically on both Linux and macOS, and the manual-install hint in `build-and-import-image.sh` is now just a fallback/safety net.

## Verification

- On a Linux VM/container without docker preinstalled: run `sudo ./setup-env.sh -u $USER`, then confirm `docker buildx version` succeeds.
- On a Linux box with docker already installed but no buildx plugin: run setup again and confirm the new standalone check installs `docker-buildx-plugin` without reinstalling docker-ce.
- On macOS: run `sudo ./setup-env.sh -u $USER`, confirm Colima/docker install proceeds as before and `docker buildx version` succeeds afterward.
- Run `src/utils/build-and-import-image.sh` end-to-end on both platforms to confirm the existing buildx-missing error path is no longer hit.

Ismael-Sallami:
Following your message about publishing images instead of building them locally, I published the OpenSPP image to my own Docker Hub and tested this PR against it. The short version is that this PR is what makes that direction work.

I built it from a clean clone of the pinned tag and pushed it with the script we already have:

`ismaelyz23/openspp:19.0`, `linux/amd64`, digest `sha256:9b466a7e9c4a65a07484cfe2e6062838d993859149fb15b461d08c410a9179ca`

Then I ran the same command with the same config file twice, changing only the code.

On `dev`:

```
==> Deploying OpenSPP
ERROR  Image ismaelyz23/openspp:19.0 not found in the cluster (build-only).
ERROR  Build and import it with:  src/utils/build-and-import-image.sh -n ismaelyz23/openspp -t 19.0 ...
```

It aborts, and it asks you to build an image that is already published and pullable.

On this branch:

```
INFO   Image ismaelyz23/openspp:19.0 is not loaded but can be pulled. Letting Kubernetes pull it.
    Creating TLS secret 'openspp-tls' (4 SANs)              [  ok  ]
    Deploying OpenSPP Helm chart                            [  ok  ]
    Waiting for PostGIS and Odoo to become Ready            [  ok  ]
    Smoke test (/web/login + OpenSPP module)                [  ok  ]
```

The kubelet pulled it in 33.9s (636 MB) and the image digest in the running pod is the same one the registry reports. Three pods 1/1, no sudo, no `k3s ctr import`, nothing built locally.

One detail that makes the test stricter: the config file I used comes from `dev`, so it has no `OPENSPP_BUILD_IF_MISSING` key. It defaults to false, so building was never an option. The deploy either pulled the image or stopped.

Testing this also found something to fix in the PR itself, and it is pushed now. The pullable check ran `docker manifest inspect`, and when `docker` was missing it read that as "not pullable". The node is what pulls the image, not the client, so on a machine that only has kubectl the deploy stopped even though the image was reachable. That is exactly the remote-cluster case you asked me to test. It now tells three cases apart: pullable, the registry does not have it, and no tool here to ask with. The last one lets Kubernetes pull and says so in the log.

On buildx: with the image published the end user does not build anything, so `docker buildx` is only needed by whoever publishes. That is the same reason you gave for holding off. I would still like to keep it in its own sub-task so this PR can go in.

One thing I noticed while testing and did not change here: in `deployer.sh` the `cleanup_openspp` call runs before `deploy_openspp`, and the image check is the first thing inside it. With the default `redeploy=true`, a failing image check deletes the namespace first and aborts after. I used `-r false` for these runs. I can open a separate issue for it if you want.

Ismael-Sallami:
Two updates since the comment above.

A second commit fixes the wording that assumed building is the only way to get the image in: the prerequisites list in `docs/OPENSPP.md` and the comment above the image keys in `config.ini`. No behaviour change. The same wording in `docs/BUILDING-IMAGES.md` and in the chart's `values.yaml` is in #172, which already touches both files.

The tag is multi-arch now: `ismaelyz23/openspp:19.0`, digest `sha256:b962b3e7ce76b89e407773053860914794f8ed9541648e385273b909231fb12f`, with `linux/amd64` and `linux/arm64` in it. The amd64 entry is the same `sha256:9b466a7e...` the run above deployed, so that evidence still covers it. arm64 was built natively on an OCI Ampere box in about 6 minutes.

I did not change the image values in `config.ini`, they still say `ghcr.io/openmf/openspp` and `19.0`: mine is a personal repo and should not be the project default. When you re-tag, the keys are `OPENSPP_IMAGE_REPOSITORY` and `OPENSPP_IMAGE_TAG` under `[openspp]`. One thing that is easy to miss: the current default names ghcr.io, and a Docker Hub image would be `openmf/openspp` with no host in front, because a first segment without a dot means Docker Hub.

tdaly61 (approved):

PR #174, opened 2026-08-07
GAZ-327: Add docs/MIFOSX.md describing the MifosX deployment

kanishksingh23:
**Summary**

- Document each MifosX component Gazelle deploys, how to use it, and where the upstream documentation lives: Fineract, the Angular web app, the shared PostgreSQL, and the three integrated extension modules (Pentaho reporting, Flowable workflow engine, credit bureau integration). SMS & Messaging and Loan Assessment are covered as planned, with the reasons they are not yet deployed.

- Includes the MifosX product architecture annotated to show which components are integrated, planned, or out of scope for Gazelle, and a section on the pattern for adding a new MifosX module.

- Known limitations are stated explicitly, notably that Pentaho per-tenant routing needs a fixed plugin release, and that the workflow and credit bureau images are temporary until published under openMF.

- Also links the new document from the deployment guide.


docs/MIFOSX.md
tdaly61, line 1: This is a great start Kanishk and really helpful. I have some suggestions and will inline them but can you add a table of contents please.

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+
+Gazelle deploys MifosX as its own app (`-a mifosx`) into the `mifosx` namespace, tracking the MifosX **25.12.25** release line — Apache Fineract 1.14.0, the Angular web app, and PostgreSQL 16.6. Beyond the core platform, Gazelle integrates three of MifosX's extension modules: Pentaho reporting, the Flowable workflow engine, and credit bureau integration.
+
+This document describes each component Gazelle deploys, how to use it, and where the deeper upstream documentation lives.
+
+---
+
+## Deployment Method
+
+Unlike OpenG2P and Payment Hub EE, MifosX is deployed from **vendored Kubernetes manifests** rather than Helm charts. Every manifest lives in `src/deployer/manifests/mifosx/` — no repo clone, no chart repository, no submodule.
```
tdaly61: can you remove the references to other deployments here please, just make it simple that MifosX is deployed from manifests and the manifests live in / under `src/deployer/manifests/mifosx/

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+
+## How It Fits into Mifos Gazelle
+
+The MifosX product architecture, showing the modules that make up the platform:
+
+![Mifos X Product and Roadmap architecture](mifosx-images/mifos-x-product-architecture.png)
+
+The same architecture annotated to show which components Mifos Gazelle actually integrates:
+
+```
```
tdaly61, line 58: I thnk for this section you might consider just leaving MySQL and kafka out , mention somewhere they are not used by MifosX Gazelle deployments

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+| Port | `8081` |
+| Health | `GET /actuator/health` |
+| API base | `/api/v1` |
+| Database | `mifos_flowable` on the shared PostgreSQL |
+| Fineract tenant | `greenbank` |
+
+BPMN definitions are auto-deployed from the image at startup and Flowable creates its own schema, so nothing needs restoring. Two initContainers gate startup: `wait-postgres` and `wait-fineract`.
+
+**Using it.** The service has no ingress, so port-forward it:
+
```
tdaly61, line 208: Do e need an ingress here to avoid the port-forward and make things simple ??

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+```bash
+curl -u mifos:password -X POST http://localhost:8081/api/v1/auth/authenticate
+curl -u mifos:password -X POST http://localhost:8081/api/v1/workflow/client-onboarding/start \
+  -H 'Content-Type: application/json' -d '{ ... }'
+```
+
+**Why it runs on PostgreSQL.** The module was MySQL-only, while the move to PostgreSQL left Gazelle's shared infrastructure with no MySQL for MifosX to use. Adding a dedicated MySQL pod would have cost roughly 512Mi against the 16GB budget, and borrowing Payment Hub EE's MySQL would have stopped MifosX deploying standalone. Instead the module was made database-agnostic at source: the PostgreSQL driver was added *alongside* MySQL, the datasource made environment-overridable with MySQL still the default, and the hardcoded MySQL dialect removed so Hibernate auto-detects. One image runs on either database and the environment picks — so the change is additive and upstreamable rather than a fork. It is low-risk because the module defines no JPA entities of its own: the database holds only Flowable's tables, and Flowable ships per-database DDL.
```
tdaly61: It runs on Postgres in-line with the move of MifosX to standardise on Postgres ..I think that is all you need to say here (we will put something in the release notes too , but that is outside of this doc)

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+
+| | |
+|---|---|
+| Port | `8081` |
+| Health | `GET /credit-bureaus` |
+| Database | `creditbureau` on the shared PostgreSQL |
+
+Gazelle runs it with `CDC_MOCK_ENABLED=true` — the default bureau target is Círculo de Crédito, and no real bureau credentials are configured. Two initContainers gate startup: `ensure-creditbureau-db` creates the database if it does not exist, and `wait-for-fineract` blocks until the core API answers.
+
+**Using it.** No ingress, so port-forward:
```
tdaly61: again don't we need an ingress , to make things simpler here for the users.  Oh and to be clear I mean an ingress added to the manifests so that this doc basically just asks the user to browse or curl to a URL

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+
+---
+
+## Known Limitations
+
+- **Pentaho per-tenant routing needs a fixed plugin release.** The published plugin resolves its datasource in a way that can return another tenant's data. The fix is a two-line change, merged upstream into `openMF/mifos-reporting-plugin` ([PR #513](https://github.com/openMF/mifos-reporting-plugin/pull/513), `pentaho` branch), but **no fixed release has been published yet** and Gazelle installs the published artifact. Until a fixed release ships, treat multi-tenant report output as unreliable. Closing this needs either a new plugin release or a temporary class swap in the initContainer.
+- **Financial reports render empty on the seeded tenants.** Reports run, but totals come out zero because the seeded tenants have no posted general-ledger entries — Gazelle's demo data is payment-oriented (clients and savings accounts for transfers) rather than loan- and GL-heavy. This is a demo-data gap, not a reporting fault.
```
tdaly61: this is a todo to fix though with our data loading ...isn't it ? To be clear, we can leave this in now as it is correct now but we ought to try and remedy this for release.

docs/MIFOSX.md
```diff
@@ -0,0 +1,335 @@
...
+- **The Workflow Engine and Credit Bureau images are temporary.** Neither project publishes a container image, so both are built from source and pushed to a personal DockerHub namespace. Both pins should move to `openMF` images once those are published.
```
tdaly61: please see my long slack message about publishing  images from the script and testing for image existence. These are script / devops changes but will affect this doc too

tdaly61 (changes requested):

tdaly61 (approved):

PR #179, opened 2026-08-08
GAZ-331: move OpenSPP2 to the 2026.08 release

Ismael-Sallami:
## Summary

Gazelle pins OpenSPP2 to the first v2 release. Upstream has published two since, and this moves both pins to the second one, `2026.08`. The first of the two redesigned two registry flows and that redesign broke modules built on top of OpenSPP, which `2026.08` reverts. It also stops the container printing its generated Odoo configuration, which contains the database password, into the pod log on every start.

Value-only change: 15 lines across 7 files, no logic touched.

## Ticket

GAZ-331 (sub-task of gaz 192). Cut off current `dev`.

## What is included

**`config/config.ini` and `src/deployer/openspp.sh`**: `OPENSPP_SOURCE_REF` and `OPENSPP_IMAGE_TAG`, and the deployer defaults that have to match them. **`src/deployer/helm/openspp/`**: chart `appVersion` and image tag, with the chart `version` bumped. **`docs/OPENSPP.md`, `docs/BUILDING-IMAGES.md` and `tests/openspp/circleci-job.yml`**: the same two values where they are written out.

Both keys move together on purpose. The tag `19.0` names the Odoo platform, not the OpenSPP release, and both releases run on Odoo 19, so it cannot tell them apart. `openspp_ensure_image()` compares that tag against the image names already on the node, without opening any image, so moving only the source reference would leave a machine that has built before deploying the old code, with a green smoke test.

## How to test

```bash
./run.sh -m deploy -a openspp
bash tests/openspp/smoke.sh
bash demos/openspp/run_demo.sh
bash demos/openspp/run_demo.sh
```

On a machine that has built before, remove the old image first, otherwise there is nothing to prove. To confirm the deploy is running this release and not a cached image, `spp_program_geofence` exists only in it:

```bash
kubectl exec -n openspp statefulset/openspp-postgis -- psql -U odoo -d openspp -tAc \
  "SELECT count(*) FROM ir_module_module WHERE name='spp_program_geofence'"
```

Observed: smoke 4/4 on amd64 and on arm64; the marker returns 1; the demo credits 5/5 in MifosX from an empty database with the cycle ending `distributed`, and a second run pays 0 and still passes. The payment connector, which is not in this diff but is the module most exposed to that release note, installs on the new database with its linter clean and its test suite green. On arm64 the geospatial database image has to be pointed at a multi-arch build, because the official one is amd64 only.

## Checklist

- [x] Both keys updated consistently in the configuration and in the deployer defaults.
- [x] Chart `appVersion` updated and chart `version` bumped.
- [x] Guides and the parked CI job updated; no file still names the old source tag.
- [x] `helm lint` passes and `helm template` renders `2026.08` on the application image.
- [x] Clean deploy and smoke test green on amd64 and on arm64.
- [x] Agri demo green from an empty database and idempotent on a second run.

## AI usage

`claude-code` was used to help explain concepts, test the code, and with documentation. All output was reviewed, tested and adjusted by the author.


src/deployer/openspp.sh
```diff
@@ -158,13 +158,13 @@ openspp_check_prerequisites() {
     OPENSPP_NAMESPACE="${OPENSPP_NAMESPACE:-openspp}"
     OPENSPP_RELEASE_NAME="${OPENSPP_RELEASE_NAME:-openspp}"
     OPENSPP_IMAGE_REPOSITORY="${OPENSPP_IMAGE_REPOSITORY:-ghcr.io/openmf/openspp}"
-    OPENSPP_IMAGE_TAG="${OPENSPP_IMAGE_TAG:-19.0}"
+    OPENSPP_IMAGE_TAG="${OPENSPP_IMAGE_TAG:-2026.08}"
```
tdaly61: This release tag is sort of buried in here , it would be better if it were in a yaml file or in the config.ini as the next person trying to update has to search in multiple places

tests/openspp/circleci-job.yml
```diff
@@ -25,10 +25,10 @@ commands:
       - run:
           name: Build and import the OpenSPP2 image (build-only, not published)
           command: |
-            git clone --branch v19.0.2.0.0 --depth 1 https://github.com/OpenSPP/OpenSPP2.git repos/OpenSPP2
-            docker build --target production -t ghcr.io/openmf/openspp:19.0 \
+            git clone --branch 2026.08 --depth 1 https://github.com/OpenSPP/OpenSPP2.git repos/OpenSPP2
```
tdaly61: Same thing , lets have all tags and versions declared outside of .sh or .py files

tdaly61 (dismissed):

Ismael-Sallami:
Thanks @tdaly61, both done, and the problem was a bit worse than it looked from the diff.

The tag was already in `config.ini`. What you spotted in `openspp.sh` was a second copy of it, written as a default (`${OPENSPP_IMAGE_TAG:-2026.08}`), which quietly wins whenever the key is missing. The two copies were already drifting: `config.ini` ships `OPENSPP_BUILD_IF_MISSING = true` and the script defaulted it to `false`.

So instead of moving the tag, I stopped declaring it twice:

- `openspp.sh` now requires six keys and stops naming the missing one rather than guessing. The line I drew is "nothing that can go stale": the two images with their tags, and the source repository with its git reference. Namespace, release name and paths keep their defaults.
- `circleci-job.yml` reads the values with `crudini`, which the setup step above it already installs.
- The chart uses what `helm create` scaffolds, `tag: ""` plus `| default .Chart.AppVersion`, so the tag is written once, in `Chart.yaml`, and the chart still installs on its own.
- `docs/OPENSPP.md` now says in one place that moving to a new release is those two keys, and why both are needed.

There is no version left in any `.sh` or `.py` file. The description above is from the first version of this branch: it is 10 files now, and it does add a small piece of logic, the check.

**Testing.** Rebased on `dev` first, 7 files conflicted, keeping the wording and image names and only the version from here. The check exits naming the key both when it is empty and when it is missing. Rendered with the same `--set` the deployer uses, this chart and the one on `dev` give identical output apart from the chart label. `circleci config validate` and `helm lint` pass. Deployed on a VM running the full stack: pods Ready, smoke 4/4, and the image still `2026.08` with no tag in `values.yaml`.

DavidH-1 (approved):
issues raised by @tdaly61 fixed

PR #182, opened 2026-08-11
GAZ-338: lend against the farmer registry in the agri demo

Ismael-Sallami:
## Summary

Gazelle shows OpenSPP paying a subsidy into MifosX, and the registry behind that demo is empty: the fixtures carry the crop, the land size and the zone, and the loader reads none of them. A farmer registry that holds only names and phone numbers cannot be used for the thing a farmer registry exists for.

This adds the other half. The same households get their farm recorded, OpenSPP's own scoring engine turns those attributes into a credit decision, the farmer consents to that decision being shared, and MifosX opens and disburses the loan it justifies. The subsidy demo is untouched: this is a second act on the same data, in its own script, and the same farmer ends up holding a government subsidy and a bank loan in the same account.

The decision is not an `if` inside a script. It is a scoring model configured in OpenSPP with five indicators read straight from the registry, three score bands, and a result that keeps the value of every attribute, what each one contributed and the version of the model. That record is what reaches the bank and what can be audited afterwards.

```mermaid
flowchart LR
    R["OpenSPP farmer registry"] --> S["spp_scoring: five indicators, three bands"]
    S --> C["spp_consent: signed before the data leaves"]
    C --> D{"Which rail"}
    D -- direct --> F["Fineract bluebank: into the account that received the subsidy"]
    D -- workflow --> W["Mifos workflow engine: Credit Assessment step"]
    W --> G["Fineract greenbank: paid out of the branch"]
    F --> E["The loan carries the score, the band, the collateral and the purpose"]
    G --> E
```

## Ticket

GAZ-338 (sub-task of gaz 192). Cut off current `dev`.

## What is included

**`src/utils/openspp/openspp-agri-credit.py`** (new): reads the enrolled farmers and their farm out of OpenSPP, makes sure the scorecard exists and is active, scores everyone, records consent, and then opens the loan. It reports the decision as a table before anything reaches the bank, so what the registry decided is visible whether or not the loan goes through.

**The decision.** Five indicators, each read by field path from the registry: farm size, years farming, land tenure, livestock held, and whether the land is productive, which is required. They add up to a score out of 100, three bands map that score to terms, and the amount is `100 + 150 per hectare` capped by the band. With the fixtures as they ship, the five farmers come out at 90, 90, 80, 52 and 42, so the demo shows three bands rather than five identical loans. Indicators read the registry by field path and not through the engine's expression language, because its expression context carries only the common registrant fields and a formula would quietly score zero on every farm.

**Consent.** A subsidy is the registry paying its own beneficiary; a loan sends personal data to a third party. So a consent record is created first, naming who signs, which organisation receives the data, for which purposes, which categories of data and until when. A farmer with no consent on record is skipped and the run says so. The consent is filed on both the household and the person who signs, which is where the registry's own interface shows it.

**Two rails, chosen at run time.** `direct` opens the loan against Fineract in the tenant where the farmers already bank, so the disbursement lands in the same savings account that received the subsidy. `workflow` drives the Mifos workflow engine's own BPMN loan origination process, whose Credit Assessment step is where the score from the registry enters the bank's process. `auto` uses the workflow engine when it answers and falls back to Fineract saying so. Either can be forced with `ORIGINATION`.

**What the loan carries.** The registry identifier as the client's external id, a datatable row with the score, the band, the model version, the hectares, the crop and the consent, the land parcel as collateral, and a loan purpose. So the loan does not only exist: it says what it was granted on.

**`src/utils/openspp/verify-loan-in-mifosx.py`** (new): for each farmer, finds the loan, checks it is active, and **recomputes the amount from the evidence stored on the loan**. A loan whose amount does not match the farm it was granted on is a failure and not a pass.

**`src/utils/openspp/load-openspp-agri-data.py`**: writes the farm behind each household onto the partner it already created, using the delegation the registry model provides, plus a land parcel, a crop activity with the crop as a vocabulary code, and the head of household as a registrant of their own. Additive: a fixture without the agricultural columns loads exactly as before, which is what keeps the subsidy demo working.

**`demos/openspp/run_credit_demo.sh`** (new) and `docs/OPENSPP.md`: the orchestrator for the second act, and the section that explains what it demonstrates, how the decision is made, the two rails and their limits. Two screenshots go with it, the scoring results in OpenSPP and the same decision carried on the loan in MifosX, in the folder the guide already uses. `run_demo.sh` is not touched.

## How to test

```bash
bash demos/openspp/run_demo.sh
ORIGINATION=direct bash demos/openspp/run_credit_demo.sh
ORIGINATION=workflow bash demos/openspp/run_credit_demo.sh
bash tests/openspp/smoke.sh
```

The workflow rail needs `workflow.<domain>` in `/etc/hosts`. A machine set up before the engine had an ingress does not have it, because only the environment setup writes that file.

This was validated with #169, #179 and #180 applied on top of `dev`, because that is the tree this work will live in once they merge, and applying them is the only way to exercise the whole flow. On `dev` as it stands the configured image repository does not resolve, so the deploy falls back to building the image and the first run costs about 30 minutes. Nothing here depends on those three, though: the code and the guide in this pull request apply to `dev` on their own.

Validated on a cluster rebuilt from nothing: the applications removed, the stack redeployed from current `dev` with the three open pull requests applied, and OpenSPP deployed from the published image without building anything. The smoke test passes its four checks. The subsidy demo credits 5/5 through OpenSPP's native connector from an empty database, so the shared loader is safe. The direct rail then opens five loans of 580, 400, 325, 165 and 132 in the tenant where the farmers already bank, and the workflow rail opens five more through its own process in the tenant fixed at deploy time. Ten loans, all active, and the amounts match on both rails because the registry decides them and not the rail. A second run on each rail opens nothing and still passes, and a run cut short is carried to the end rather than abandoned. Read back from Fineract, one loan reports purpose `Agricultural inputs`, its linked savings account, and the registry evidence: registrant `openspp-12`, score 90, band A, 3.2 hectares of rice.

Then the same on arm64, on an Ampere machine: the deploy pulls the published image and builds nothing, the database runs native, and both rails open five loans each for the same five amounts. The only difference is one configuration key for the database image, because the official PostGIS image publishes no arm64 build.

Not covered: the workflow engine does not forward a linked savings account to Fineract, so on that rail the loan is paid out with no destination account and the farmer's savings balance does not move, which the guide states rather than works around. That rail also needs its hostname to resolve, and only the environment setup writes that file, so a machine prepared before the engine had an ingress has to add the line by hand. And the subsidy demo was not re-run on the arm64 machine, because it needs Payment Hub and the switch and that run deployed neither.

## Checklist

- [x] Registry holds the farm behind each household: size, tenure, experience, crop as a vocabulary code and a land parcel.
- [x] Scoring model created, activated and scoring five farmers into three bands, with the per-indicator breakdown stored.
- [x] Consent recorded for every farmer and visible on the household and on the person who signs.
- [x] Direct rail opens, secures, approves and disburses five loans into the account that received the subsidy.
- [x] Workflow rail creates, assesses and approves five loans through its own process.
- [x] Loan carries external id, registry evidence, collateral and purpose, checked through the API.
- [x] Second run on each rail opens nothing and passes; an interrupted run is resumed to the end.
- [x] Subsidy demo green end to end on the same base after the loader change.
- [x] Guide updated with the demo, the decision, the two rails and their limits.
- [x] Guide shows both ends: the scoring results in OpenSPP and the same decision carried on the loan in MifosX.
- [x] Both rails run on arm64 too, with the same five amounts and one configuration key for the database image.

## AI usage

`claude-code` was used to help explain concepts, test the code, and with documentation. All output was reviewed, tested and adjusted by the author.


DavidH-1 (approved):
LGTM

PR #184, opened 2026-08-13
GAZ-339: make the agri demo survive dropped connections and a busy Odoo

Ismael-Sallami:
## Summary
Two failures of the agri demo, both found by running it and both only visible under timing.

The demo keeps one XML-RPC connection to OpenSPP for the whole run, but between one payment and the
next it spends minutes polling MifosX without touching OpenSPP. The ingress closes an idle connection
at 75 seconds, so the next call dies and takes the run with it. Now the connection is rebuilt and the
call retried once when it arrives closed.

The demo also installs the payment module when the image carries it but has not installed it yet.
Odoo refuses module operations while a scheduled action is running, and the demo read that refusal as
"the install started and the connection dropped", so it polled a state nobody was going to change and
then died blaming slowness. Now the two are told apart: a refusal is retried, a dropped install is
polled.

## Ticket
GAZ-339, sub-task of gaz 192. Cut off current `dev`.

Both fixes are here because they are the same mistake in the same demo: treating a transport failure
as if the server were slow.

## What is included
- `src/utils/openspp/openspp-agri-demo.py`: the call closure reconnects and retries once on a closed
  connection.
- `src/utils/openspp/load-openspp-agri-data.py`: the module installer retries Odoo's refusal inside
  the budget it already had, and names the scheduled action when it runs out.

## How it works
`xmlrpc.client` already retries a request once when the cached connection turns out to be cold, but
only for `RemoteDisconnected` and for `ECONNRESET`, `ECONNABORTED` and `EPIPE`. Over plain HTTP a
closed connection lands in that list and the retry hides it; over TLS the same close arrives as
`ssl.SSLEOFError` and propagates. So moving the demo to the ingress both introduced a proxy that
closes idle connections and took that retry out of play. The fix sits in the call closure, which
covers the connector and the bridge without touching either payment loop.

Retrying is safe: these errors are raised on send, with the socket already closed, so the request
never reached Odoo. The only calls that create work, `spp.payment create` and `prepare_payment`, are
never the first after a long wait because a read always comes first.

For the install, Odoo runs `SELECT FROM ir_cron FOR UPDATE` before any module operation and raises
`Odoo is currently processing a scheduled action` when a cron holds the table. The install has not
started at that point, so polling waits for nothing. A fresh database fires every cron at once, which
is why this lands on the first run after a clean deploy.

## Checklist
- [x] Both failures forced against current `dev` first, to check the forcing reproduces them.
- [x] Both rails covered: the fix is in the connection, not in either payment loop.
- [x] The install keeps its 60 x 5 s budget, so a run cannot take longer than before.
- [x] Full demo green on both rails, and a re-run stays idempotent.
- [x] No new dependency: `http.client` is standard library and `ssl` was already imported.

## AI usage
`claude-code` was used to help explain concepts, test the code, and with documentation. All output
was reviewed, tested and adjusted by the author.


DavidH-1 (approved):
LGTM

PR #186, opened 2026-08-16
GAZ-306: Enrich MifosX demo data so reports render meaningful data

kanishksingh23:
## Summary

Enriches the MifosX demo seed data so the Pentaho reports render meaningful data instead of empty tables. Previously, each demo tenant had a single client and no transaction/accounting history, so most reports came up empty.

- **Differentiated tenants**: greenbank = payer (3 accounts), bluebank = payee (12 account holders), redbank = 1.
- **Transaction history**: accounts are opened ~120 days back and receive a deterministic series of deposits/withdrawals, so the savings/transaction reports show a real timeline with running balances instead of a single same-day deposit.
- **CASH accounting**: the savings product is created with a minimal chart of accounts mapped for CASH accounting, so every deposit/withdrawal posts GL journal entries. This populates the Trial Balance with real, balanced figures.
- **Bug fix**: `check_client_exists_by_mobile` relied on Fineract's `/clients?mobileNo=` filter, which isn't applied server-side, so every new MSISDN matched the first client, and only one client was ever created per tenant. Now filtered client-side.
- Regenerated `config/fineract-db-dump-final.sql` with the enriched data.

Verified on a fresh `./run.sh -m deploy -a mifosx`: greenbank 3 / bluebank 12 (unique names) / redbank 1 restored correctly, and Client Listing, Savings Transactions, and Trial Balance render meaningful, balanced data.

## Note

- The first 6 bluebank MSISDNs are the OpenG2P beneficiary numbers, preserved unchanged and in the same order, so the OpenG2P demo data is unaffected.
- Loan reports remain empty by design, these tenants are payment/savings-oriented (no loans).
- Balance Sheet / Income Statement still mis-render their asset section (assets show 0, cash lands in retained earnings). This is a separate Pentaho report-template/plugin issue with those reports' deeply nested subreports, not a demo-data problem: the underlying GL data is correct, as the balanced Trial Balance shows. Tracked separately.


DavidH-1 (approved):
Lgtm

PR #193, opened 2026-08-22
GAZ-346: Workflow Engine make it usable & demoable in Gazelle

kanishksingh23:
## Summary

Makes the Workflow Engine module demo-able end-to-end.

- Pin the fixed image `kanishksingh23/mifos-workflow:21082026`. It includes the delegate ID cast fix ([openMF/mifos-workflow#77](https://github.com/openMF/mifos-workflow/pull/77)); the previous `07082026` image threw a `ClassCastException` at the assign-staff step, so a client onboarding could start but never complete.
- Add `src/utils/demo-workflow.sh`, a one-command end-to-end demo: authenticate → ensure a loan officer exists → start onboarding → approve the verify task (assign staff + activate) → confirm the client is Active in Fineract. Reaches the service via `kubectl port-forward`, so no domain setup is needed.
- Document it in `docs/MIFOSX.md` (Workflow Engine section): the one-command demo plus how to complete a task by hand.

Verified on k3s: `./src/utils/demo-workflow.sh` onboards a client all the way to `Active`.

## Note

- The image is a personal-namespace tag for now (`kanishksingh23/…`), same as the module's existing pin swap to `openmf/mifos-workflow` once[ #77](https://github.com/openMF/mifos-workflow/pull/77) is merged and the image is republished under the org.
- Greenbank ships no staff, so the demo creates a loan officer if none exists (assign-staff needs one).


src/utils/demo-workflow.sh
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env bash
```
tdaly61, line 1: hi @kanishksingh23 , does this run on a remote cluster ?

kanishksingh23: Hey @tdaly61, yes it runs on a remote cluster. The script reaches the services via kubectl port-forward, which tunnels through the API server (the same connection kubectl uses), so it works wherever your kubeconfig points; no ingress/DNS/etc. hosts setup is needed, and only the API server needs to be reachable (not node ports or the ingress).

tdaly61 (approved):

PR #194, opened 2026-08-23
GAZ-353: Fix OOM pods and make-payment failure on full deployment

kanishksingh23:
## Summary
Fixes OOM pods and make-payment.sh script not completing during a full deployment. Two causes:

- **Zeebe OOM**: the PaymentHub Zeebe broker was OOM-killed at its own cgroup memory limit (1920Mi), independent of node RAM. Raised the limit to 2560Mi.
- **Fineract probes too aggressive**: under full-stack load, Fineract's health endpoint answers slower than the tight liveness/readiness timeouts (period 1s / timeout 1–2s), so it was pulled out of its Service (ingress 503 → make-payment "no payer") and restarted by liveness. Loosened the probe period/timeout/failureThreshold.

## Note
- Validated end-to-end on a 15GB VM via a staged deploy: Fineract ready in 171s / 0 restarts, no OOM, all per-DFSP BPMN processes registered, and make-payment.sh completes (HTTP 200). 

- Reported on a 32 GB ARM64 VM and reproduced on a 15 GB AMD64 VM too; both causes are RAM/architecture-independent (a cgroup pod-limit OOM and load-sensitive probe timeouts), so they apply regardless of node size.

**SS**
Screenshot 2026-08-23 at 5 21 07 PM


tdaly61 (approved):
looks good thanks @kanishksingh23

PR #197, opened 2026-08-27
GAZ-345: Add preflight to demo-credit-bureau.sh so it fails clearly if MifosX is not deployed

kanishksingh23:
**Summary**

Adds a preflight to `src/utils/demo-credit-bureau.sh` so it fails immediately with a clear message when MifosX isn't deployed, instead of backgrounding a port-forward and hanging before a vague error. Matches the preflight already in `demo-message-gateway.sh` and `demo-loan-module.sh`.

## Note

- Adds a `fail()` helper plus checks that the `mifosx` namespace and `credit-bureau` service exist.
- Verified: `./demo-credit-bureau.sh -n doesnotexist` now exits with "`Is MifosX deployed? Run: .run.sh -m deploy -a mifosx"
`


kanishksingh23:
Hey @tdaly61, while working on the sms module's script, I noticed it failed silently, so I added a clear message in case it fails. I did the same thing in this module also.

tdaly61 (approved):

PR #200, opened 2026-09-07
GAZ-309: Add MifosX Upgrade runbook and VERSIONS.md

kanishksingh23:
## Summary

Nothing is recorded about which image versions make up the MifosX deployment or what to do
when a new MifosX release arrives.

- **`src/deployer/manifests/mifosx/VERSIONS.md`** is now the single source of truth
  for the pins: the six application images, the helper images the initContainers
  use, the Pentaho plugin, and the shared PostgreSQL from the infra chart. Every row
  names the manifest it is pinned in. No `latest` tags.
- **`docs/UPGRADE-RUNBOOK.md`** is the checklist for adopting a future release: read
  the MifosX release notes, move `VERSIONS.md` and the manifest pins together,
  rebuild the three modules Gazelle builds from source, redeploy, verify, raise the PR.
- **`docs/MIFOSX.md`** now links both from its Version Pins section, which is what
  that section was always meant to reference, the links were left out only because
  neither file existed yet.
 - All four `demo.sh` scripts were verified against this branch and passed: a client
onboarded to Active by the Workflow Engine, a mock credit report, an SMS reaching
DELIVERED through the Dummy provider, and a loan assessment row written via Kafka.

## Note

- **This PR also corrects six stale claims in `docs/MIFOSX.md`.** It already edits
  that file to restore the two links, so the fixes ride along rather than needing a
  ticket of their own. The one worth reading is the message gateway: the document
  said the gateway "accepts and records requests but cannot deliver", which is wrong
  twice over: without a provider bridge, `POST /sms` returns a 500, and with one the
  built-in Dummy provider does simulate delivery. Registering a bridge is the step
  that makes the module usable, and it appeared nowhere. The others are drift: a
  `{ ... }` placeholder where a runnable request belonged, two of four module
  databases listed, "three modules" where there are five, and a `/etc/hosts` note
  naming two of four module hostnames.

### DIFF
--- a/docs/MIFOSX.md
+++ b/docs/MIFOSX.md
@@ -109,7 +109,7 @@ Components deployed into the `mifosx` namespace:
 | Fineract | `fineract-server` pod | `openmf/fineract:1.14.0` | Core banking engine and API |
 | Web App | `web-app` pod + ingress | `openmf/web-app:dev-10d24b8` | Angular user interface |
 | Reports | *inside* `fineract-server` | plugin staged by initContainer | Pentaho formatted reporting |
-| Workflow Engine | `mifos-workflow` pod + ingress | `kanishksingh23/mifos-workflow:07082026` | Flowable BPMN process orchestration |
+| Workflow Engine | `mifos-workflow` pod + ingress | `kanishksingh23/mifos-workflow:21082026` | Flowable BPMN process orchestration |
 | Credit Bureau | `credit-bureau` pod + ingress | `kanishksingh23/mifos-credit-bureau:21082026` | Credit bureau integration |
 | SMS & Messaging | `message-gateway` pod + ingress | `openmf/message-gateway:dev-51abedb` | SMS and email delivery |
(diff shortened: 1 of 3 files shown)


tdaly61 (approved):
yeah this is good @kanishksingh23 thanks heaps.  Just bear in mind I need to create openMF versions of these images prior to release but this simply changes the base and will mean very very little other changes are required.

One note: I plan that we will release Gazelle 2.1.0 without having all the MifosX release and publishing fixes in (there are still 2 tickets I need to ensure you tagged in ..but I need to find em again, Victor lodged them to I believe fix the release process here).  What you have in the upgrade guide is excellent and will really help coordinate the overall process fix.   thanks again

kanishksingh23:
Thank you @tdaly61, for reviewing and merging the PR. About the Victor ticket I think you are probably talking about this one https://mifosforge.jira.com/browse/WEB-1023. If so I have gone through it once and saw someone raised PR for this but yeah I need to give a look again.

PR #201, opened 2026-09-09
GAZ-48: Add the MifosX work to the v2.1.0 release notes

kanishksingh23:
**Summary**

- The v2.1.0 draft release changes MifosX from a partially deployed Fineract and web app only to the full suite with all five extension modules.

- Major New Features leads with that, then covers each module, the four one-command demo scripts, docs/MIFOSX.md, and the new VERSIONS.md and upgrade runbook. Notable Changes covers the 25.12.25 re-pin, per-module ingresses, the image-existence check, enriched demo data, and the ARM64, OOM, and setup-env fixes.
 
- Adds a Known Limitations section, because release notes are where someone planning a real deployment finds out what does not work yet: Pentaho reports do not run until a fixed plugin release is cut upstream, three module images are built from source and pinned to a personal DockerHub namespace, no live SMS provider is configured, and the loan assessment module only sees the default tenant.

### DIFF
--- a/docs/RELEASE-NOTES.md
+++ b/docs/RELEASE-NOTES.md
@@ -2,6 +2,16 @@
 
 ## Major New Features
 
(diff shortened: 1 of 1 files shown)


tdaly61 (approved):
yeah nice!

